PR #11, opened 2020-07-22
Add User class

dash102:
User class and appropriate getters/mutators.
Java check fails because the Event class doesn't exist yet, but will be solved when issue #3 is resolved.

### DIFF
--- a/project/src/main/java/com/google/sps/data/User.java
+++ b/project/src/main/java/com/google/sps/data/User.java
@@ -188,4 +188,9 @@ public final class User {
   public Builder toBuilder() {
     return new Builder(this.name, this.email).mergeFrom(this);
   }
+
+  @Override
+  public String toString() {
+    return String.format("User:\t%s\nID:\t%d\nE-mail:\t%s", this.name, this.userId, this.email);
+  }
 }
--- a/project/src/main/java/com/google/sps/data/User.java
+++ b/project/src/main/java/com/google/sps/data/User.java
@@ -30,27 +30,27 @@ public final class User {
       this.email = email;
     }
 
-    public Builder interests(Set<String> interests) {
+    public Builder setInterests(Set<String> interests) {
       this.interests = interests;
       return this;
     }
 
-    public Builder skills(Set<String> skills) {
+    public Builder setSkills(Set<String> skills) {
       this.skills = skills;
       return this;
     }
 
-    public Builder eventsHosting(Set<Event> eventsHosting) {
+    public Builder setEventsHosting(Set<Event> eventsHosting) {
       this.eventsHosting = eventsHosting;
       return this;
     }
 
-    public Builder eventsParticipating(Set<Event> eventsParticipating) {
+    public Builder setEventsParticipating(Set<Event> eventsParticipating) {
       this.eventsParticipating = eventsParticipating;
       return this;
     }
 
-    public Builder eventsVolunteering(Set<Event> eventsVolunteering) {
+    public Builder setEventsVolunteering(Set<Event> eventsVolunteering) {
       this.eventsVolunteering = eventsVolunteering;
       return this;
     }
--- a/project/src/main/java/com/google/sps/data/User.java
+++ b/project/src/main/java/com/google/sps/data/User.java
@@ -85,7 +85,7 @@ public final class User {
   private User(Builder builder) {
     this.name = builder.name;
     this.email = builder.email;
-    
+
     this.interests = builder.interests;
     this.skills = builder.skills;
     this.eventsHosting = builder.eventsHosting;
--- a/project/src/main/java/com/google/sps/data/User.java
+++ b/project/src/main/java/com/google/sps/data/User.java
@@ -1,10 +1,10 @@
-package com.google.sps;
+package com.google.sps.data;
 
 import java.util.HashSet;
 import java.util.Set;
 
 public final class User {
-  private final long userId;
+  private final long userId = 10000000L;
   private String name;
   private String email;
   private Set<String> interests;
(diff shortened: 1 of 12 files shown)


project/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,92 @@
+package com.google.sps;
```
nikhitashanker, line 1: Should be package com.google.sps.data;

dash102: Whoops, done

dash102:
Seems like AutoValue wouldn't apply in this situation, as the class needs to be mutable.

project/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,92 @@
...
+   * Creates a new user.
+   *
+   * @param name The human-readable name for the user. Must be non-null.
+   * @param email The e-mail with which the user is reachable. Must be non-null.
+   */
+  public User(String name, String email) {
+    this.userId = 10000000L;
+    this.name = name;
+    this.email = email;
+    this.interests = new HashSet<>();
```
liangqigong, line 26: I think you can move all these to line 10-14?

project/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,92 @@
...
+  private Set<Event> eventsVolunteering;
+
+  /**
+   * Creates a new user.
+   *
+   * @param name The human-readable name for the user. Must be non-null.
+   * @param email The e-mail with which the user is reachable. Must be non-null.
+   */
+  public User(String name, String email) {
+    this.userId = 10000000L;
```
liangqigong, line 23: Remove?

project/src/main/java/com/google/sps/data/User.java
dash102: Changed due to structure revamp

project/src/main/java/com/google/sps/data/User.java
dash102: Changed due to structure revamp

dash102:
Even though the tests for the User builder functionality aren't necessary, I would definitely appreciate any input anyone has on them :)

project/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,127 @@
...
+    private Set<Event> eventsHosting = new HashSet<>();
+    private Set<Event> eventsParticipating = new HashSet<>();
+    private Set<Event> eventsVolunteering = new HashSet<>();
+
+    public Builder(String name, String email) {
+      this.name = name;
+      this.email = email;
+    }
+
+    public Builder interests(Set<String> interests) {
```
liangqigong, line 33: setInterests. And for the following functions.

project/src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final String SKILL_TO_ADD = "Juggling";
+
+  // The following fields will be instantiated with meaningful Events after the Events class is
+  // written to follow the Builder pattern.
+  private static final Set<Event> EVENTS_HOSTING = Collections.emptySet();
+  private static final Set<Event> EVENTS_PARTICIPATING = Collections.emptySet();
+  private static final Set<Event> EVENTS_VOLUNTEERING = Collections.emptySet();
+
+  @Test
+  public void getNameAfterBuild() {
```
liangqigong, line 34: I think you don't need so many tests since using Builder to create a User class is one behavior. We should test behavior, not function.

So I think you only need 2 tests:
  1. Use Builder to create one class, and verify all fields are correctly set.
  2. Use toBuilder to transfer a User to User.Builder, and verify all fields are transferred.

project/src/main/java/com/google/sps/data/User.java
dash102: Done

project/src/test/java/com/google/sps/UserTest.java
dash102: Done, also added ones to check mutation since there are add/remove methods now

dash102:
Ready for review

project/src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+/** */
+@RunWith(JUnit4.class)
+public final class UserTest {
+  private static final String NAME = "Bob Smith";
+  private static final String EMAIL = "bobsmith@example.com";
+  private static final Set<String> INTERESTS = new HashSet<>(Arrays.asList("Conservation", "Food"));
+  private static final Set<String> NEW_INTERESTS =
+      new HashSet<>(Arrays.asList("Conservation", "Food", "Music"));
+  private static final Set<String> SKILLS = new HashSet<>(Arrays.asList("Cooking"));
```
nikhitashanker, line 23: I think you can use ImmutableList.of here from guava.

project/src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,196 @@
...
+    public Builder removeEventVolunteering(Event eventVolunteering) {
+      this.eventsVolunteering.remove(eventVolunteering);
+      return this;
+    }
+
+    public User build() {
+      return new User(this);
+    }
+
+    public Builder mergeFrom(User other) {
```
liangqigong, line 122: Does this need to be public?

project/src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,78 @@
...
+    User expectedUser =
+        new User.Builder(NAME, EMAIL)
+            .setInterests(INTERESTS)
+            .setSkills(SKILLS)
+            .setEventsHosting(EVENTS_HOSTING)
+            .setEventsParticipating(EVENTS_PARTICIPATING)
+            .setEventsVolunteering(EVENTS_VOLUNTEERING)
+            .build();
+
+    User actualUser = expectedUser.toBuilder().build();
```
liangqigong, line 66: Empty line below

project/src/test/java/com/google/sps/UserTest.java
```diff
@@ -0,0 +1,78 @@
...
+    Assert.assertTrue(EqualsBuilder.reflectionEquals(expectedUser, actualUser));
+  }
+
+  @Test
+  public void addToExistingInterests() {
```
liangqigong, line 71: The Test is a little bit confusing.

I think in user.getInterests(), there should be INTERESTS + NEW_INTEREST?

I guess you can just remove '.setInterests(INTERESTS)'.

dash102: If I remove .setInterests(INTERESTS), wouldn't I have empty interests?
The idea of this test is to make sure addInterest() works properly, so I made a user with 2 interests and added a new 3rd one.

project/src/test/java/com/google/sps/UserTest.java
dash102: Done

project/src/main/java/com/google/sps/data/User.java
dash102: Ah, thanks for catching that - thanks

project/src/test/java/com/google/sps/UserTest.java
dash102: Done

dash102:
Ready for review.

project/src/test/java/com/google/sps/UserTest.java
liangqigong: But why user.getInterests() equals to NEW_INTERESTS, shouldn't it equal to INTERESTS + NEW_INTEREST?

dash102: Is there a good way in Java to concatenate on the fly like that? I wasn't able to find one that didn't either mutate the original array or create a copy (and the latter is essentially what my code is doing now)

dash102: It has been passing the tests and the logic still makes sense to me, is this because it's bad practice?

I updated the naming of the variables to make it less confusing.

PR #12, opened 2020-07-22
Add Event class

yicyao:
#3 
Adds Event class for hardcoded events for Prototype.
Eventually, in the MVP, constructor and all setters and getters will be taken directly from the database.

**Build fails because this Class relies on User and VolunteerOpportunity classes which are not in this PR

### DIFF
--- a/project/src/main/webapp/sps/data/Event.java
+++ b/project/src/main/webapp/sps/data/Event.java
@@ -1,0 +1,96 @@
+import java.util.Date;
+import java.util.Random; 
+import java.util.Set;
+
+/** Class containing Event object 
+Setters for variables that user can change about event
+*/
+public class Event {
+  private final int eventId;
+  private String name;
+  private String description;
+  private Set<String> labels;
+  private String location;
+  private Date date;
+  private Set<VolunteeringOpportunity> opportunities;
+  private Set<User> attendees;
+  private User host;
+
+  Random rand = new Random(); 
+
+  /** TO DO (MVP): Add Event to Event db*/
+  public Event(String name, String description, Set<String> labels, String location, Date date, 
+  Set<VolunteeringOpportunity> opportunities, Set<User> attendees, User host) {
+    this.eventId = rand.nextInt(1000000); //this will later be a unique number based on the database id
+    this.name = name;
+    this.description = description;
+    this.labels = labels;
+    this.location = location;
+    this.date = date;
+    this.opportunities = opportunities;
+    this.attendees = attendees;
+    this.host = host;
+  }
+
+  /** TO DO (MVP) for all getters: get from Event db*/
+  /** TO DO (MVP) for all setters: set in Event db*/
+  public int getID() {
+    return eventId;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String newName) {
+    this.name = newName;
+  }
+
+  public String getDescription() {
+    return description;
+  }
+
+  public void setDescription(String newDescription) {
+    this.description = newDescription;
+  }
+
+  public Set<String> getLabels() {
+    return labels;
+  }
+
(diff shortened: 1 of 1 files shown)


project/src/main/webapp/sps/data/Event.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public void setDate(Date newDate) {
+    this.date = newDate;
+  }
+
+  public Set<VolunteeringOpportunity> getOpportunities() {
+    return opportunities;
+  }
+
+  public Set<User> getAttendees() {
```
nikhitashanker, line 89: Missing addAttendee(User attendee).

project/src/main/webapp/sps/data/Event.java
```diff
@@ -0,0 +1,96 @@
...
+  private String location;
+  private Date date;
+  private Set<VolunteeringOpportunity> opportunities;
+  private Set<User> attendees;
+  private User host;
+
+  Random rand = new Random(); 
+
+  /** TO DO (MVP): Add Event to Event db*/
+  public Event(String name, String description, Set<String> labels, String location, Date date,
```
dash102, line 22: To me it makes more sense to just initialize attendees as empty in the constructor, and then add attendees as people sign up

project/src/main/webapp/sps/data/Event.java
```diff
@@ -0,0 +1,96 @@
...
+  private String name;
+  private String description;
+  private Set<String> labels;
+  private String location;
+  private Date date;
+  private Set<VolunteeringOpportunity> opportunities;
+  private Set<User> attendees;
+  private User host;
+
+  Random rand = new Random();
```
hlz03, line 19: Is rand used anywhere else? if not you can set in constructor.

nikhitashanker:
Maybe fix the Java Presubmit failing build.

project/src/main/webapp/sps/data/Event.java
yicyao: my reasoning for setting it not in the constructor is that many Event objects will be created and I dont want to reinitialize rand each time

liangqigong: Then I guess you should make it static final?

project/src/main/webapp/sps/data/Event.java
```diff
@@ -0,0 +1,96 @@
...
+  private Set<VolunteeringOpportunity> opportunities;
+  private Set<User> attendees;
+  private User host;
+
+  Random rand = new Random(); 
+
+  /** TO DO (MVP): Add Event to Event db*/
+  public Event(String name, String description, Set<String> labels, String location, Date date, 
+  Set<VolunteeringOpportunity> opportunities, Set<User> attendees, User host) {
+    this.eventId = rand.nextInt(1000000); //this will later be a unique number based on the database id
```
liangqigong, line 24: Why do you need a random number here?

yicyao: Technically the number wont be random, it should be a unique number as the event ID which will come from the unique hash of the  event in the database -- i just put a random large number for now to prevent ID collisions
This is all hardcoded for the prototype which will change next week -- however I can also change this to make it a guaranteed unique int
(We wont create more than 10 events for the prototype)

project/src/main/webapp/sps/data/Event.java
yicyao: thanks done!

project/src/main/webapp/sps/data/Event.java
yicyao: resolved

yicyao:
I'm not sure how this got automatically merged when I merged PR #15 . Github will not let me unmerge -- Event class with Builder and tests in new PR

Issue #7, opened 2020-07-22
Add UI to create an event


yicyao:
(hardcoded, without connecting to DB)

PR #15, opened 2020-07-22
Add Event Creation UI form

yicyao:
PR #7 
Add Event Creation UI form based on decided upon UI -- hardcoded form for prototype

Looks like:
14RYS2MxMAs

TO DO: add header to top of page when header is pushed to master


project/src/main/webapp/create-event.html
```diff
@@ -0,0 +1,94 @@
...
+      />
+    <script
+      src="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/js/bootstrap.min.js"
+      integrity="sha384-OgVRvuATP1z7JjHLkuOU7Xw704+h835Lr+6QL9UvYjZE3Ipu6Tp75j7Bh/kR0JKI"
+      crossorigin="anonymous"
+      ></script>
+  </head>
+  <body>
+    <br><br>
+    <h3 style="text-align:center" > Create an Event </h3>
```
dash102, line 22: I think these styling statements should go in the css so it's easier to search for and edit later

yicyao: i wanted to stylize these inline because I dont want to override styles like <h3> if someone else uses them and wants them default / to look like something else -- let me know if that makes sense 

same with like bootstrap buttons etc

dash102: We can totally discuss this in chat - I personally think that we should keep styles in a CSS file with IDs/classes as needed. As for whether we want to do a 1:1 with CSS/HTML files, that might be a best practices question?

yicyao: resolved. done

project/src/main/webapp/create-event.html
```diff
@@ -0,0 +1,95 @@
...
+          </div>
+        </div>
+      </div>
+      <br>
+      <div class="inline">
+        <text> Tags: </text>
+        <div class="container">
+          <span class="badge badge-primary text-medium"> Tag 1</span>
+          <span class="badge badge-primary text-medium"> Tag 2</span>
+          <span class="badge badge-secondary text-medium"> + Tag</span>
```
dash102, line 58: Are the leading spaces to provide some kind of spacing? 
If so it might be better for that to go in CSS padding - if not, then it might not be necessary

yicyao: good catch, deleted (done)

PR #39, opened 2020-07-29
[Need some feedback] Add injection example for database.

nikhitashanker:
- Add a way for tests to test business logic without making database calls
- By default servlet uses init to instantiate InstanceDatabaseWrapper 
- In the test we can use setWrapper to change DatabaseWrapper to TestDatabaseWrapper
-  Use mockito to mock the servlet request response
- Dependency injection followed from Guice 101: Topic 0: No Guice
- When ServletExampleWithInjectionTest is run, "Not actual class invocation" will be printed from TestDatabaseWrapper's insertUser function and the InstanceDatabaseWrapper insertUser function will not be called.
- TODO: Fix style.


project/src/test/java/com/google/sps/ServletExampleWithInjectionTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.Before;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(JUnit4.class)
+public class ServletExampleWithInjectionTest extends Mockito {
+  private final TestInstanceDatabaseWrapper wrapper = new TestInstanceDatabaseWrapper();
```
dash102, line 23: Maybe declare this as its interface instead so we aren't depending on any `TestInstanceDatabaseWrapper`-specific methods - this way we are actually guaranteeing that we're using and testing "real" methods that the real `DatabaseWrapper` has and not something that only exists in test world

project/pom.xml
```diff
@@ -16,11 +16,28 @@
     <failOnMissingWebXml>false</failOnMissingWebXml>
   </properties>
   <dependencies>
+      <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
     <dependency>
```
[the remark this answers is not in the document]
nikhitashanker: Done - thanks.

project/src/test/java/com/google/sps/ServletExampleWithInjectionTest.java
nikhitashanker: I think if its an interface we possibly would not be able to do the injection properly because interface implies there is no implementation of any methods?

dash102: Not that we would instantiate it as in interface (since interfaces can't do that), but `private final DatabaseWrapper wrapper = new TestInstanceDatabaseWrapper()'` where it's of the interface type instead would (I think) guarantee that we can't use methods that are only existent in the TestInstanceDatabaseWrapper class

I don't think it would affect injection, but maybe wait and see what liangqi and hailun say

nikhitashanker: Done - Oh I see, yes that makes sense now. Thank you.

nikhitashanker:
Not ready to merge but need some feedback if this is the correct method.

project/src/main/java/com/google/sps/servlets/ServletExampleWithInjection.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
```
yicyao, line 17: nit: alphabetize

project/src/main/java/com/google/sps/servlets/ServletExampleWithInjection.java
```diff
@@ -0,0 +1,59 @@
...
+
+/** Servlet that returns the login status of the user. */
+@WebServlet("/it-servlet-example")
+public class ServletExampleWithInjection extends HttpServlet {
```
yicyao, line 21: maybe name class better than 'ServletExampleWithInjection' to what the Servlet actually does, and comment that this is the injection example... same with test name

project/src/test/java/com/google/sps/ServletExampleWithInjectionTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(JUnit4.class)
+public class ServletExampleWithInjectionTest extends Mockito {
+  private final DatabaseWrapper wrapper = new TestDatabaseWrapper();
+  private final ServletExampleWithInjection servlet = new ServletExampleWithInjection();
+
+  @Before
+  public void setUp() throws Exception {
+    servlet.setWrapper(wrapper);
```
liangqigong, line 28: The thought is correct, but I don't think you particularly use injection here.

I think in the example I gave you, it uses a Module (PetTestModule) to config which Pet to use. So you don't need to call setPet function, the Pet instance is injected to the class.

Could you maybe try to use a Module to control which Wrapper to use? Let me know if you have trouble to find how to do so.

nikhitashanker: Thanks, so my thinking was that since this is a Servlet, it is not recommended to use the constructor to inject a parameter (I saw some resources like this https://stackoverflow.com/questions/25868944/servlet-does-not-have-a-default-constructor). That is why I opted to use a set function.

Let me know if you think I can still try to use a module to control which wrapper to use.

project/src/main/java/com/google/sps/utilities/InstanceDatabaseWrapper.java
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Given a user, insert a row with all available fields into the DB
+   *
+   * @param user the user to be inserted; user's ID field should not exist in DB
+   */
+  public void insertUser(User user) {
+    SpannerOptions options = SpannerOptions.newBuilder().build();
+    Spanner spanner = options.getService();
+    DatabaseId db = DatabaseId.of(options.getProjectId(), instanceID, databaseID);
+    DatabaseClient dbClient = spanner.getDatabaseClient(db);
```
hlz03, line 31: Can this be a class variable and initialize in constructor? So the all the actions can just call dbClient.

project/src/test/java/com/google/sps/ServletExampleWithInjectionTest.java
liangqigong: I found some questions on Injection in Servlet, could you take a look and let me know if any of them works?

https://www.google.com/search?q=servlet+injection+site:stackoverflow.com&sxsrf=ALeKk02c8uEN_mQb3c4rRUvBOykLjtMEtw:1596217023377&sa=X&ved=2ahUKEwjU96Llg_jqAhXPvZ4KHZmHB5MQrQIoBHoECAYQBQ&biw=1120&bih=611&dpr=3

nikhitashanker: Will do, thanks.

project/src/main/java/com/google/sps/utilities/InstanceDatabaseWrapper.java
nikhitashanker: Ack - I will try to make a new PR from master since this code is already on master.

nikhitashanker: The question @dash102 and I have about this is how to call spanner.close() after all operations are complete within the DatabaseWrapper class if we move this to the constructor.
A possible way is to create helper function closeSpanner() in DatabaseWrapper and force all servlets to call this method at the end but this does not seem clean.

nikhitashanker:
Closed by https://github.com/googleinterns/step188-2020/pull/57

PR #38, opened 2020-07-29
[Draft] Event DB Logic

yicyao:
[Not ready for review]
DB logic for adding and updating events when the user
Tests needed

User submits form to create new event --> backend sends event to db --> redirects to new event page with newly created event details

#22 #32

PR #40, opened 2020-07-29
Modify User ID generation

dash102:


dash102:
Ready for review. Build fails because it depends on a small Event.java change that will be done in #38

dash102:
Closing because I am no longer keying Users by ID

PR #51, opened 2020-08-03
[2/Q] Add servlet to retrieve single volunteering opportunity.

nikhitashanker:
- Add servlet and database wrapper method to retrieve single volunteering opportunity.

### DIFF
--- a/project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
@@ -23,7 +23,7 @@ public class VolunteeringOpportunityDataServlet extends HttpServlet {
     String opportunityId = request.getParameter(OPPORTUNITY_ID);
 
     Optional<VolunteeringOpportunity> opportunity =
-        databaseWrapper.getVolunteeringOpportunityForOppportunityId(opportunityId);
+        databaseWrapper.getVolunteeringOpportunityByOppportunityId(opportunityId);
 
     if (opportunity.isEmpty()) {
       response.setContentType("text/html;");
--- a/project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
@@ -18,6 +18,14 @@ public class VolunteeringOpportunityDataServlet extends HttpServlet {
   private static final DatabaseWrapper databaseWrapper =
       new DatabaseWrapper(DatabaseConstants.INSTANCE_ID, DatabaseConstants.DATABASE_ID);
 
+  /**
+   * Queries database for opportunity with opportunity ID given in the request parameter
+   * and writes opportunity to the response.
+   *
+   * @param request servlet request
+   * @param response servlet response.
+   * @throws IOException if Input/Output error occurs
+   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String opportunityId = request.getParameter(OPPORTUNITY_ID);
--- a/project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
@@ -19,11 +19,11 @@ public class VolunteeringOpportunityDataServlet extends HttpServlet {
       new DatabaseWrapper(DatabaseConstants.INSTANCE_ID, DatabaseConstants.DATABASE_ID);
 
   /**
-   * Queries database for opportunity with opportunity ID given in the request parameter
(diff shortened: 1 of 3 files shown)


project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
```diff
@@ -123,4 +128,45 @@ public void insertVolunteeringOpportunity(VolunteeringOpportunity opportunity) {
...
     return mutations;
   }
+
+  /**
+   * Given an opportunityId, retrieve the corresponding volunteering opportunity
+   *
+   * @param opportunityId opportunityId of the opportunity to retrieve
+   * @return volunteering opportunity wrapped in a {@link Optional}
+   */
+  public Optional<VolunteeringOpportunity> getVolunteeringOpportunityForOppportunityId(
```
hlz03, line 138: getVolunteeringOpportunityByOpportunityId.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
```diff
@@ -123,4 +128,45 @@ public void insertVolunteeringOpportunity(VolunteeringOpportunity opportunity) {
...
+      String opportunityId) {
+    SpannerOptions options = SpannerOptions.newBuilder().build();
+    Spanner spanner = options.getService();
+    DatabaseId db = DatabaseId.of(options.getProjectId(), instanceId, databaseId);
+    DatabaseClient dbClient = spanner.getDatabaseClient(db);
```
hlz03, line 143: Are you going to use injection for DatabaseWrapper? if so, these operations and spanner.close does not need to be here.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
nikhitashanker: Done.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
nikhitashanker: Ack, I can make a new PR to do this from master since it would effect other functions as well. So if I understand correctly, I can refactor this by creating a class called DatabaseClient.java and passing in DatabaseClient in the constructor on the DatabaseWrapper class. And I can put all these operations in the constructor and put spanner.close() in a DatabaseClient function that can be called by DatabaseWrapper?

nikhitashanker: Created a PR for this (no injection) on master and will make change here after approval.

project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet to get volunteering opportunity given opportunityId */
+@WebServlet("/volunteering-opportunity-data")
+public class VolunteeringOpportunityDataServlet extends HttpServlet {
+  private static final String OPPORTUNITY_ID = "opportunity-id";
+  private static final DatabaseWrapper databaseWrapper =
+      new DatabaseWrapper(DatabaseConstants.INSTANCE_ID, DatabaseConstants.DATABASE_ID);
+
+  @Override
```
liangqigong, line 21: Just realized that you should also leave a comment for the function, since the name doesn't explain the function.

```
7.3 Where Javadoc is used
At the minimum, Javadoc is present for every public class, and every public or protected member of such a class, with a few exceptions noted below.

Additional Javadoc content may also be present, as explained in Section 7.3.4, Non-required Javadoc.

7.3.1 Exception: self-explanatory members
Javadoc is optional for "simple, obvious" members like getFoo(), in cases where there really and truly is nothing else worthwhile to say but "Returns the foo".

Important: it is not appropriate to cite this exception to justify omitting relevant information that a typical reader might need to know. For example, for a method named getCanonicalName, don't omit its documentation (with the rationale that it would say only /** Returns the canonical name. */) if a typical reader may have no idea what the term "canonical name" means!

7.3.2 Exception: overrides
Javadoc is not always present on a method that overrides a supertype method. See advice for documenting override methods.

7.3.3 Exception: the google3/javatests/... tree
Javadoc is not required for code under the google3/javatests/... tree. But it is not forbidden, and can be useful especially for functional test classes and for methods and classes used by multiple tests.

7.3.4 Non-required Javadoc
Other classes and members have Javadoc as needed or desired.

Whenever an implementation comment would be used to define the overall purpose or behavior of a class or member, that comment is written as Javadoc instead (using /**).

Non-required Javadoc is not strictly required to follow the formatting rules of Sections 7.1.2, 7.1.3, and 7.2, though it is of course recommended.
```

nikhitashanker: Done - added comments to all public servlet functions.

PR #57, opened 2020-08-03
[1/X] Remove duplication in DatabaseWrapper.java and using a single database connection.

nikhitashanker:
- Remove duplication 
- Change format to follow https://github.com/GoogleCloudPlatform/java-docs-samples/blob/c7c0de9512a60e9728fb112f185d09bf19321c4b/appengine-java11/spanner/src/main/java/com/example/appengine/spanner/SpannerTasks.java so there is only 1 time a database connection is opened.


nikhitashanker:
Ready for review.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
```diff
@@ -25,14 +19,9 @@ public DatabaseWrapper(String instanceId, String databaseId) {
...
-    SpannerOptions options = SpannerOptions.newBuilder().build();
-    Spanner spanner = options.getService();
-    DatabaseId db = DatabaseId.of(options.getProjectId(), instanceId, databaseId);
-    DatabaseClient dbClient = spanner.getDatabaseClient(db);
-
     List<Mutation> mutations = getMutationsFromBuilder(newInsertBuilderFromUser(), user);
-    dbClient.write(mutations);
-    spanner.close();
+    databaseService.getDatabaseClient().write(mutations);
+    databaseService.getSpanner().close();
```
hlz03, line 24: No need to close the Spanner here?

This should be called when you turn down the whole server.

nikhitashanker: Done. I see - thanks.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
```diff
@@ -82,4 +64,9 @@ public void updateUser(User user) {
     mutations.add(builder.build());
     return mutations;
   }
+
+  /** Close the Spanner database connection. */
```
liangqigong, line 68: Just curious, if one servlet close the Spanner, does it affect other Servlets to use the Wrapper?

nikhitashanker: After checking documentation, this will only close the spanner instance created in the constructor in DatabaseService. Each servlet has its own spanner instance as it will call new DatabaseWrapper(new DatabaseService()).

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
```diff
@@ -0,0 +1,30 @@
+package com.google.sps.utilities;
+
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+
+public class WrapperDatabaseService implements DatabaseService {
+  public static final String INSTANCE_ID = "step-188-instance";
```
dash102, line 9: Use the DatabaseConstants class that's on master rn

liangqigong: +1

nikhitashanker: Done - thanks.

project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
```diff
@@ -23,13 +24,14 @@
...
 @WebServlet("/create-event")
 public class EventCreationServlet extends HttpServlet {
-  private static final DatabaseWrapper dbWrapper = new DatabaseWrapper("step-188-instance", "event-organizer-db");
-
   /** Returns event details from database */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String eventId = request.getParameter("eventId");
+
+    DatabaseWrapper dbWrapper = new DatabaseWrapper(new WrapperDatabaseService());
```
yicyao, line 32: Could this use the same dbWrapper as the doPost() in line 63? like private static final or not?

nikhitashanker: We cannot because this is called only when the servlet is initialized (runs only once). So if we initialize using private static final and close it in doGet(), then the next time we do a GET, the connection that came from when the servlet is initialized will be closed.

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
```diff
@@ -0,0 +1,29 @@
+package com.google.sps.utilities;
+
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+
```
yicyao, line 7: add javadoc or quick comment

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
```diff
@@ -300,33 +266,33 @@ public void updateVolunteeringOpportunity(VolunteeringOpportunity opportunity) {
...
-        long numSpotsLeft = resultSet.getLong(2);
-        Set<String> requiredSkills =
-            resultSet.getStringList(3).stream().collect(Collectors.toSet());
-        results.add(
-            new VolunteeringOpportunity.Builder(eventId, name, numSpotsLeft)
-                .setOpportunityId(opportunityId)
-                .setRequiredSkills(requiredSkills)
-                .build());
-      }
+    try (ResultSet resultSet =
```
yicyao, line 276: except?

project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
yicyao: Ok makes sense ty for the clarification!

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
nikhitashanker: Ack - Google example does not having anything here, strange but I think I will keep it.

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
nikhitashanker: Done.

yicyao:
LGTM

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
```diff
@@ -0,0 +1,30 @@
...
+/** Class that initializes spanner and database client. */
+public class WrapperDatabaseService implements DatabaseService {
```
hlz03, line 9: DatabaseServiceImpl instead of WrapperDatabaseService.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
```diff
@@ -20,21 +20,15 @@
...
-  private static final SpannerOptions options = SpannerOptions.newBuilder().build();
-  private static final Spanner spanner = options.getService();
-  private final DatabaseId db = DatabaseId.of(options.getProjectId(), instanceId, databaseId);
-  private final DatabaseClient dbClient = spanner.getDatabaseClient(db);
 
-  public DatabaseWrapper(String instanceId, String databaseId) {
-    this.instanceId = instanceId;
-    this.databaseId = databaseId;
+  public DatabaseWrapper(DatabaseService databaseService) {
+    this.databaseService = databaseService;
```
hlz03, line 31: why not also have a private final databaseClient = databaserService.getDatabaseClient() here?

nikhitashanker: I think we cannot because this is called only when the servlet is initialized (runs only once). So if we initialize using private static final and close it in doGet(), then the next time we do a GET, the connection that came from when the servlet is initialized will be closed.

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
```diff
@@ -0,0 +1,30 @@
...
+  private final Spanner spanner;
+  private final DatabaseClient databaseClient;
+
+  public WrapperDatabaseService() {
+    SpannerOptions options = SpannerOptions.newBuilder().build();
+    spanner = options.getService();
+    DatabaseId database =
+        DatabaseId.of(options.getProjectId(), DatabaseConstants.INSTANCE_ID, DatabaseConstants.DATABASE_ID);
+    databaseClient = spanner.getDatabaseClient(database);
```
liangqigong, line 18: Could you move these to an initialize function?

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
nikhitashanker: Done.

project/src/main/java/com/google/sps/utilities/DatabaseService.java
```diff
@@ -0,0 +1,10 @@
+package com.google.sps.utilities;
+
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.Spanner;
+
+public interface DatabaseService {
```
liangqigong, line 6: I think it would be better if the interface provides query, write, close instead of spanner and client. So writer and reader of DatabaseWrapper don't need to know the details of interaction of DB.

project/src/main/java/com/google/sps/utilities/DatabaseWrapper.java
nikhitashanker: Ack - I can change it, I was using this example for dependency injection where they call owner.getPet() in the main function. I was thinking owner is like DatabaseWrapper and Pet is like databaseService. https://docs.google.com/document/d/1gAf3i91d1PSmNr329BziJbNBc81JbfxdhXYnSivjxNg/edit#heading=h.ny5w2jug54cu

project/src/main/java/com/google/sps/utilities/SpannerClient.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    return databaseClient;
+  }
+
+  @Override
+  public void contextInitialized(ServletContextEvent event) {
+    if (event != null) {
+      sc = event.getServletContext();
+      if (INSTANCE_ID == null) {
+        INSTANCE_ID = sc.getInitParameter("SPANNER_INSTANCE");
```
dash102, line 91: I would write a `getValueOrDefault(T value, T default)` kind of thing to do all the null checks

project/src/main/java/com/google/sps/utilities/DatabaseService.java
nikhitashanker: Ack, no longer applies - thank you for reviewing.

project/src/main/java/com/google/sps/utilities/WrapperDatabaseService.java
nikhitashanker: Ack, no longer applies - may be useful in future PRs.

project/src/main/java/com/google/sps/utilities/SpannerClient.java
nikhitashanker: Done. Added ternary operator due to logic change.

nikhitashanker:
[Ready for review]

PR #60, opened 2020-08-04
Versioning fixes

dash102:
- Removes duplicate dependency
- Updates Guava version to fix Dependabot's vulnerability notice
- Updates method uses to be compatible with Java 8


project/src/main/java/com/google/sps/servlets/VolunteeringOpportunityDataServlet.java
```diff
@@ -33,7 +33,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     Optional<VolunteeringOpportunity> opportunity =
         databaseWrapper.getVolunteeringOpportunityByOppportunityId(opportunityId);
 
-    if (opportunity.isEmpty()) {
+    if (!opportunity.isPresent()) {
```
liangqigong, line 36: Let change this to opportunity.isPresent() and swap the blocks? I think it will be more straightforward for readers..

dash102: Done

PR #65, opened 2020-08-04
Add email field to login status.

nikhitashanker:

PR #67, opened 2020-08-05
Restructure linking for logged-in/-out users

dash102:
Cleans up some naming and structuring of code related to redirecting users upon login/logout
Dynamically sets the purpose of the logo depending on if the user is logged in or out


dash102:
Javascript linter is complaining that JSDoc is incomplete, but the function titles seem descriptive enough so that any docs would be redundant

project/src/main/webapp/script.js
```diff
@@ -58,3 +56,9 @@ async function addLogoutUrlToButton() {
   const logoutPrompt = document.getElementById('logout-prompt');
   logoutPrompt.href = logoutUrl;
 }
+
+async function isLoggedIn() {
```
yicyao, line 60: same function isLoggedIn() in 2 diff js files (script.js and index.js). Is it good practice to only have one copy but in the HTML files import both scripts, or have a diff copy of each function per js file? @hlz03 @liangqigong

project/src/main/webapp/index.js
```diff
@@ -1,14 +1,31 @@
...
+async function addLoginUrlToElement(id) {
+  const loginUrl = await getLoginUrl();
+  const element = document.getElementById(id);
+  element.href = loginUrl;
+}
+
+async function isLoggedIn() {
+  const response = await fetch('/login-status');
+  const loginStatus = await response.json();
+  return loginStatus.isLoggedIn;
```
nikhitashanker, line 24: change to (loginStatus === "LOGGED_IN") due to change in PR #65

dash102: Done

project/src/main/webapp/script.js
dash102: Done

PR #69, opened 2020-08-05
Fix input tag styling

dash102:
Moves tag-related styling to a shared CSS file to avoid repeated code
Correctly adds in bootstrap-tagsinput to allow for dynamically-updated tags


project/src/main/webapp/create-event.html
```diff
@@ -73,12 +76,21 @@ <h3>Create an Event</h3>
...
       <div class="inline">
-        <text> Tags: </text>
-        <div class="container">
-          <span class="badge badge-primary text-medium"> Tag 1</span>
-          <span class="badge badge-primary text-medium"> Tag 2</span>
-          <span class="badge badge-secondary text-medium"> + Tag</span>
-        </div>
+        <span> Tags: </span>
+        <label for="interests">Interests</label>
+        <select
```
yicyao, line 81: Everything LGTM. Question for later PRs, if we want the options for the labels, do we just fill < select > out with option tags? Eg. https://developer.snapappointments.com/bootstrap-select/examples/

dash102: If you want options, then this plugin I'm using probably wouldn't be the move, just a normal select tag with option tags inside would work well (no "role='datainput'")

PR #74, opened 2020-08-05
[Ready for review] [2/X] Add testing for opportunity database functions.

nikhitashanker:
- setup and teardown code was adapted from Spanner Sample Code written by Stephanie Fu (@dash102) as a part of spanner setup 
[[Github Java Build is failing because Github is running database tests without database access/emulator.]]
image


project/src/main/java/com/google/sps/data/VolunteeringOpportunity.java
```diff
@@ -137,4 +137,28 @@ public Builder toBuilder() {
   public String toString() {
     return String.format("Name: %s\nNumber of spots left: %d\n", this.name, this.numSpotsLeft);
   }
+
+  @Override
```
nikhitashanker, line 141: I could not find a different way than defining an equals function. Apache's EqualsBuilder returns a strange error below. I believe it is comparing the references of the sets to see if they are equal because the error thrown does not show any differences between the two. Maybe we can add Google's AutoValue library if available to avoid this issue.
> expected Number of spots left: 240 [] []
> but was: com.google.sps.data.VolunteeringOpportunity<eventId:0883de79-17d7-49a3-a866-dbd5135062a8 Name: Performer
Number of spots left: 240 [] []

nikhitashanker:
Java Build is failing because Github is running database tests without database access/emulator.

project/src/main/java/com/google/sps/data/VolunteeringOpportunity.java
liangqigong: Could you maybe try this?
https://stackoverflow.com/questions/12147297/how-do-i-assert-equality-on-two-classes-without-an-equals-method

nikhitashanker: Done - used AssertEquals for all properties.

project/src/main/java/com/google/sps/utilities/SpannerClient.java
```diff
@@ -29,23 +29,23 @@
 
 @WebListener
 public class SpannerClient implements ServletContextListener {
-  private static String PROJECT_ID;
+  private static final String PROJECT_ID = DatabaseConstants.PROJECT_ID;
+  private static String DATABASE_ID = DatabaseConstants.DATABASE_ID;
```
hlz03, line 33: Can be removed and directly use DatabaseConstants.XX below.

project/src/test/java/com/google/sps/OpportunitySpannerTasksTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    Assert.assertEquals(actualOpportunity.getEventId(), EVENT_ID);
+    Assert.assertEquals(actualOpportunity.getName(), NAME);
+    Assert.assertEquals(actualOpportunity.getNumSpotsLeft(), NUMBER_OF_SPOTS);
+  }
+
+  @AfterClass
+  public static void tearDown() throws Exception {
+    SpannerTestTasks.cleanup();
+  }
```
hlz03, line 65: nit: We usually put setup and tearDown above, and then tests below.

project/src/main/java/com/google/sps/utilities/SpannerClient.java
nikhitashanker: Done.

PR #82, opened 2020-08-07
Add time field to event class

yicyao:
Add time field to event class and event database. 
Also added Event Tests back (I wrote a while ago but disappeared in master when merged) EDIT: moved to PR #85


yicyao:
> I don't see the Test file? (In the description you said you added the test back)

@liangqigong separated into different PR #85

PR #63, opened 2020-08-04
[Ready for review]  Make Event Editable [Backend]

yicyao:
Add edit ability to events after you create them. Redirects to the create event form with the previous form information that you can now change and write back to the db. Tests added for spanner and /create-event servlet.

Frontend in #98

PR #98, opened 2020-08-11
[Ready for review] Event editable [Frontend]

yicyao:
Add edit ability to events after you create them. Redirects to the create event form with the previous form information that you can now change and write back to the db. 
Backend in #63


project/src/main/webapp/event-details.html
```diff
@@ -3,11 +3,16 @@
   <head>
     <meta charset="UTF-8" />
     <title>STEP 188 Project</title>
+    <link rel="stylesheet" href="event-details.css" />
+    <script src="event-details.js"></script>
```
dash102, line 7: Move these scripts to be below the jquery ones

project/src/main/webapp/event-details.html
```diff
@@ -20,39 +25,33 @@
       integrity="sha384-OgVRvuATP1z7JjHLkuOU7Xw704+h835Lr+6QL9UvYjZE3Ipu6Tp75j7Bh/kR0JKI"
       crossorigin="anonymous"
     ></script>
-    <script src="script.js"></script>
-    <script src="event-details.js"></script>
   </head>
-  <body class>
+  <body onload="getEventDetails()">
```
nikhitashanker, line 29: Move this from the HTML to the javascript onload function.

PR #114, opened 2020-08-13
Fix frontend user profile tag edits

dash102:
test (4)

(this PR has a lot of lines but it's actually a lot of code movement from one file to shared.js, only some is new code)


project/src/main/webapp/shared.js
```diff
@@ -55,31 +82,124 @@ async function moveTagsFromPoolToInput(clickedAdder, tagId) {
...
+  $('#prefilled-skills').toggle();
+}
+
+function togglePrefilledInterests() {
+  $('#prefilled-interests').toggle();
+}
+
+/** Writes out relevant details to an event card */
+function populateEventContainer(event, containerId) {
+  const indexOfEventCard = 25;
```
nikhitashanker, line 131: Is there a reason this is 25? I think the variable name is not clear - maybe rename to number of event cards.

project/src/main/webapp/profile.js
```diff
@@ -1,14 +1,62 @@
...
-        $('#email').text(userEmail);
-        $('#interests').text(userInterests);
-        $('#skills').text(userSkills);
-      });
+  updateProfile();
 });
+
+async function updateProfile() {
+  const userData = await getCurrentProfileData();
+  console.log(userData);
```
nikhitashanker, line 7: remove print statement here

project/src/main/webapp/shared.js
dash102: This is the location in the object that the event card shows up, and it's always at the 25th index

project/src/main/webapp/profile.js
```diff
@@ -1,14 +1,62 @@
...
+/** Populate the user profile with name, email, interests, skills */
+function updateProfileBasics(userData) {
+  const userName = userData['name'];
+  const userEmail = userData['email'];
+  const userInterests = userData['interests'];
+  const userSkills = userData['skills'];
+  $('#name').text(userName);
+  $('#email').text(userEmail);
+  $('#interests').text(userInterests);
+  $('#skills').text(userSkills);
```
yicyao, line 27: Stylewise would it be good to not declare the const, and inline? 
eg.  $('#name').text( userData['name']); but this could just be personal preference

dash102: Yeah it's probably readable enough without two lines, I'll fix it

project/src/main/webapp/shared.js
```diff
@@ -5,30 +5,59 @@ async function isLoggedIn() {
...
-  const response = 
-      await fetch('/prefilled-information?' + new URLSearchParams({'category': informationCategory}))
+  const response =
+      await fetch(
+          `/prefilled-information?${new URLSearchParams({'category': informationCategory})}`);
   return response.json();
 }
 
-function buildTag(tagClass) {
+function buildTagWithAdder(tagClass, text) {
```
yicyao, line 14: Could you JSDoc this function? and some of the ones that are a bit less clear in this file (i think the ones in profile,js are fairly self explanatory).

project/src/main/webapp/shared.js
```diff
@@ -55,31 +82,124 @@ async function moveTagsFromPoolToInput(clickedAdder, tagId) {
...
-    });
-    $(tagId).on('itemRemoved', function(event) {
-      addTagBackToPool(event.item, tagId);
-    });
+  $(tagId).on('itemAdded', function() {
+    if ($(tagId).prevAll().length > 2) {
+      const otherId = tagId === '#interests' ? '#skills' : '#interests';
+      $(otherId).prev().remove();
+      $(tagId).prev().prev().remove();
+    }
```
yicyao, line 90: Really similar to part of profile-edit.js ' populateExisting()', could it share parts of function? (like could you build a new function for the shared lines import some of that function into that file?)

project/src/main/webapp/shared.js
```diff
@@ -55,31 +82,124 @@ async function moveTagsFromPoolToInput(clickedAdder, tagId) {
...
+    buildAsLabels(
+        `#${eventCardId} #event-card-labels`, event.labels, 'interests');
+    buildSkillsAsLabels(
+        `#${eventCardId} #event-card-labels`, event.opportunities);
+    addLinkToRegister(eventCardId);
+    addLinkToDetails(eventCardId);
+  });
+}
+
+/** Adds a hyperlink to the registration button of event card */
```
yicyao, line 156: 'Adds a hyperlink to register  to the registration button of event card ' or some comment to differentiate why &register=true vs false in addLinkToRegister() on line 157

project/src/main/webapp/shared.js
```diff
@@ -55,31 +82,124 @@ async function moveTagsFromPoolToInput(clickedAdder, tagId) {
...
+
+/**
+ * Creates the corresponding skill button labels for each volunteering
+ * opportunity
+ */
+function buildSkillsAsLabels(querySelector, opportunities) {
+  for (const opportunity of opportunities) {
+    buildAsLabels(querySelector, opportunity.requiredSkills, 'skills');
+  }
+}
```
yicyao, line 205: nit: newline

project/src/main/webapp/profile.js
dash102: Done

project/src/main/webapp/shared.js
```diff
@@ -55,31 +82,124 @@ async function moveTagsFromPoolToInput(clickedAdder, tagId) {
...
+  $(tagId).on('itemRemoved', function(event) {
+    addTagBackToPool(event.item, tagId);
+  });
+  getTagsScriptWithCallback(function() {
     $(tagId).tagsinput('refresh');
     $(tagId).tagsinput('add', $(clickedAdder).next().html());
   });
 }
 
+function getTagsScriptWithCallback(fn) {
```
nikhitashanker, line 101: maybe add javadoc, I think it is unclear what fn means.

project/src/main/webapp/shared.js
dash102: Done

project/src/main/webapp/shared.js
dash102: Done, renamed the argument to `callback` to avoid a doc

project/src/main/webapp/shared.js
dash102: done

project/src/main/webapp/shared.js
dash102: I'd like to keep the JSDoc this way so that it doesn't leak details about implementation

project/src/main/webapp/shared.js
dash102: Ah, thanks for noticing - done

PR #117, opened 2020-08-13
[Ready for review] Fix user profile to DB bug

dash102:
There was a bug where only the first item was being passed to the DB, so this PR fixes that

### DIFF
--- a/project/src/main/java/com/google/sps/utilities/TestUtils.java
+++ b/project/src/main/java/com/google/sps/utilities/TestUtils.java
@@ -15,7 +15,6 @@ public class TestUtils {
   private static final String NAME = "Performer";
   private static final int NUMBER_OF_SPOTS = 240;
   private static final String VOLUNTEER_EMAIL = "volunteer@gmail.com";
-
   private static final String USER_NAME = "Bob Smith";
   private static final Set<String> INTERESTS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList("Conservation", "Food")));
--- a/project/src/main/java/com/google/sps/utilities/TestUtils.java
+++ b/project/src/main/java/com/google/sps/utilities/TestUtils.java
@@ -2,6 +2,11 @@ package com.google.sps.utilities;
  
 import com.google.sps.data.VolunteeringOpportunity;
 import com.google.sps.data.OpportunitySignup;
+import com.google.sps.data.User;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.UUID;
 
 /* Class containing utilities for testing. */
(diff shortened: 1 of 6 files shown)


project/src/main/java/com/google/sps/utilities/TestUtils.java
```diff
@@ -10,6 +15,12 @@
   private static final String NAME = "Performer";
   private static final int NUMBER_OF_SPOTS = 240;
   private static final String VOLUNTEER_EMAIL = "volunteer@gmail.com";
+
```
liangqigong, line 18: Remove the empty line?

project/src/test/java/com/google/sps/UserProfileUpdateTest.java
```diff
@@ -0,0 +1,89 @@
...
+  @AfterClass
+  public static void tearDown() throws Exception {
+    SpannerTestTasks.cleanup();
+    authenticationHelper.tearDown();
+  }
+
+  @Test
+  public void testGetUpdatedUser() throws Exception {
+    SpannerTasks.insertOrUpdateUser(EXPECTED_USER);
+    authenticationHelper
```
liangqigong, line 63: Move this to a helper function. Maybe call it setLoggedIn or anything you prefer. Since the EMAIL and DOMAIN doesn't change any behavior of the test, so you don't want reader to be distracted.

project/src/test/java/com/google/sps/UserProfileUpdateTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+import org.springframework.mock.web.MockServletContext;
+
+/** Unit tests for DatabaseWrapper functionality related to Event class. */
+@RunWith(JUnit4.class)
+public class UserProfileUpdateTest {
+  private static final String EMAIL = "test@example.com";
+  private static final String DOMAIN = "example.com";
+  private static final User EXPECTED_USER = TestUtils.newUser(EMAIL);
```
nikhitashanker, line 34: name can be changed to USER?

project/src/main/java/com/google/sps/utilities/TestUtils.java
dash102: Done

project/src/test/java/com/google/sps/UserProfileUpdateTest.java
dash102: Done

project/src/test/java/com/google/sps/UserProfileUpdateTest.java
dash102: Done

PR #131, opened 2020-08-14
Change redirect URLs to integrate opportunity with event [back-end]

nikhitashanker:
- Add test with correct redirect URL for volunteering opportunity form servlet
- Change test for opportunity signup to have correct redirect URL
- Update servlets accordingly

### DIFF
--- a/project/src/main/java/com/google/sps/servlets/OpportunitySignupFormHandlerServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/OpportunitySignupFormHandlerServlet.java
@@ -9,7 +9,7 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
- 
+
 /** Servlet that handles submission of the form for signing up for opportunities. */
 @WebServlet("/opportunity-signup-form-handler")
 public class OpportunitySignupFormHandlerServlet extends HttpServlet {
@@ -28,7 +28,7 @@ public class OpportunitySignupFormHandlerServlet extends HttpServlet {
     String opportunityId = request.getParameter(OPPORTUNITY_ID);
     String eventId = request.getParameter(EVENT_ID);
     User user = UserServiceFactory.getUserService().getCurrentUser();
-    
+
     if (eventId == null) {
       response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Event ID not specified.");
     } else if (opportunityId == null) {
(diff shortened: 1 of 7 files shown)


project/src/test/java/com/google/sps/OpportunitySignupFormHandlerTest.java
```diff
@@ -76,20 +80,42 @@ public void testAddOpportunitySignup_LoggedInOpportunityIdNotSpecified() throws
...
 
     Mockito.verify(response)
         .sendError(HttpServletResponse.SC_BAD_REQUEST, "Opportunity ID not specified.");
   }
 
+  @Test
+  public void testAddOpportunitySignup_LoggedInEventIdNotSpecified() throws IOException {
+    VolunteeringOpportunity opportunity = TestUtils.newVolunteeringOpportunity();
+    SpannerTasks.insertVolunteeringOpportunity(opportunity);
+    authenticationHelper
```
dash102, line 96: Write a setAuthenticationHelper() kind of helper function for this to abstract away authentication details

project/src/main/java/com/google/sps/utilities/CommonUtils.java
```diff
@@ -48,6 +48,8 @@ public static String getParameter(HttpServletRequest request, String name, Strin
    */
   public static Set<String> getParameterValues(HttpServletRequest request, String name) {
     String[] values = request.getParameterValues(name);
+    if (values == null)
```
liangqigong, line 51: Always use '{}' for a code block.

```
Braces are used with if, else, for, do and while statements, even when the body is empty or contains only a single statement.
Other optional braces, such as those in a lambda expression, remain optional.
```

project/src/main/java/com/google/sps/servlets/OpportunitySignupFormHandlerServlet.java
```diff
@@ -36,7 +40,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
           new OpportunitySignup.Builder(opportunityId, user.getEmail()).build();
       SpannerTasks.insertOpportunitySignup(signup);
```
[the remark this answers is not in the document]
nikhitashanker: I think this follows google style: Line 40 starts with 4 spaces because line 39 would be too long otherwise and Line 41 starts with a 2 space indent.

project/src/main/java/com/google/sps/utilities/CommonUtils.java
nikhitashanker: Done - will keep in mind.

PR #132, opened 2020-08-14
Fix user profile UI fields

dash102:
- Makes 2 minor backend fixes that previous merges took out
- The rest of the PR is for frontend, to enforce event creation field formatting and other minor formatting


project/src/main/webapp/profile.js
```diff
@@ -46,10 +46,17 @@ async function updateUserEventsParticipating() {
...
   const userEventsVolunteering = await getUserEvents('volunteering');
   for (const eventVolunteering of userEventsVolunteering) {
-    populateEventContainer(eventVolunteering, 'events-volunteering');
+    const event = eventVolunteering.event;
+    await populateEventContainer(event, 'events-volunteering');
+    addVolunteerRole(event.eventId, eventVolunteering.opportunityName);
   }
 }
 
+function addVolunteerRole(eventId, opportunity) {
```
nikhitashanker, line 55: Add javadoc, I believe linting gives an error here otherwise?

dash102: Although the linting gives an error for some lack of JSDoc, I feel like it would produce a lot of redundancy if I add docs to functions like `addVolunteerRole` which seems like a worse idea than just a descriptive function name

PR #133, opened 2020-08-14
Add user to DB as soon as registration finishes

dash102:
Earlier, users were only added to the DB upon profile edits. This PR fixes that and makes it so that they are anonymous users in the DB upon registration.

### DIFF
--- a/project/src/main/java/com/google/sps/servlets/UserProfileUpdateServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/UserProfileUpdateServlet.java
@@ -19,41 +19,28 @@ import javax.servlet.http.HttpServletResponse;
 /** Servlet that handles user profile edits. */
 @WebServlet("/profile-update")
 public class UserProfileUpdateServlet extends HttpServlet {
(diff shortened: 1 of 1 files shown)


project/src/main/java/com/google/sps/servlets/UserProfileUpdateServlet.java
```diff
@@ -19,48 +19,36 @@
...
-              .add("eventsParticipating", CommonUtils.createJsonArray(new HashSet<>()))
-              .add("eventsVolunteering", CommonUtils.createJsonArray(new HashSet<>()))
-              .build()
-              .toString();
+      user = new User.Builder("anonymous", email).build();
+      SpannerTasks.insertOrUpdateUser(user);
     }
+    String userJson =
+        Json.createObjectBuilder()
+            .add("name", user.getName())
```
nikhitashanker, line 38: I think you should create constants for name, email, interests.

dash102: Done

PR #134, opened 2020-08-15
[2/X] Add UI to integrate opportunity with event

nikhitashanker:
- Integrate opportunity with new event UI changes and functionality
- Add button to insert opportunity (can be seen by host) on event card
image
- Add skills tags front end for insert and update opportunity [populateExistingSkills adapted from @dash102 's code]
-Add backend change to get input from tags for skills


project/src/main/webapp/event-details.js
```diff
@@ -1,32 +1,47 @@
...
 };
 
 
 /**
- * Check the current login status to update
- * isLoggedIn and currentUser.
+ * Returns the current login status.
  */
-async function checkLoginStatus() {
+async function getLoginStatus() {
```
dash102, line 14: There should be one of these in shared.js

project/src/main/webapp/event-details.js
```diff
@@ -1,32 +1,47 @@
...
+/**
+ * Should the volunteering opportunities of the event card and if the
+ * eventHost is the current user, add the edit link for each opportunity.
+ * @param {Object} opportunities
+ * @param {string} eventHost email of the host of the event
+ * @param {Object} loginStatus status of current user
+ */
+async function showVolunteeringOpportunities(
+    opportunities, eventHost, loginStatus) {
   for (const key in opportunities) {
```
dash102, line 45: Use for-of loops when you can 
https://google.github.io/styleguide/jsguide.html#features-for-loops

project/src/main/webapp/event-details.js
```diff
@@ -64,7 +82,8 @@ function getListItemForOpportunity(
 
   // If the user is logged in and the current user is the event host,
   // then show the edit link for the volunteering opportunity.
-  if (loginState === 'LOGGED_IN' && !currentUser.localeCompare(eventHost)) {
+  if (loginStatus.loginState === 'LOGGED_IN' &&
```
dash102, line 85: (this line updates if you use the shared.js function, as it abstracts away the login status value and just returns true/false)

nikhitashanker: Ack, will use in other cases but I actually need both here because I am using event host.

project/src/main/webapp/event-details.js
nikhitashanker: Done.

project/src/main/webapp/event-details.js
dash102: ah yeah - I just meant the first part - if you used the function in shared.js, the first part will just be something like if(isLoggedIn() && ...

nikhitashanker: I see - done.

project/src/main/webapp/event-details.js
nikhitashanker: Done -  added another function called getCurrentUserEmail()

project/src/main/webapp/create-volunteering-opportunity.js
```diff
@@ -1,4 +1,19 @@
...
+ * Set the opportunity form action to include query parameter for eventId.
+ */
+async function setActionToCorrectRedirectURL() {
+  const params = (new URL(document.location)).searchParams;
+  const eventId = params.get('event-id');
+
+  const opportunityForm =
+    document.getElementById('volunteering-opportunity-form');
+  opportunityForm.action =
+      `/volunteering-form-handler?event-id=${eventId}`;
```
yicyao, line 17: nit: 2 spaces instead of 4

nikhitashanker: this is when the line is too long, we go to the next line with 2 spaces - linter does not give error.

project/src/main/java/com/google/sps/servlets/VolunteeringFormHandlerServlet.java
```diff
@@ -78,4 +81,8 @@ private static void updateVolunteeringOpportunityInDB(
     // TO DO: change eventId to parameter value
     SpannerTasks.updateVolunteeringOpportunity(opportunity);
   }
+
+  private static List<String> splitAsList(String values) {
+    return Arrays.asList(values.split("\\s*,\\s*"));
+  }
```
liangqigong, line 87: Why you need to return a List? Maybe return a HashSet directly?

nikhitashanker: Done - agreed.

PR #135, opened 2020-08-15
Add Blobstore functionality (FRONTEND)

dash102:
This PR allows profile picture uploads and displays them properly


nikhitashanker:
LGTM.

project/src/main/webapp/shared.js
```diff
@@ -213,3 +213,14 @@ function buildSkillsAsLabels(querySelector, opportunities) {
     buildAsLabels(querySelector, opportunity.requiredSkills, 'skills');
   }
 }
+
+/**
+ * Adds currently-attributed profile image to logged-in user
+ * If the user has no profile image, add the default one
+ */
+async function populateExistingProfileImage() {
+  const response = await fetch('/blob-handler');
```
nikhitashanker, line 222: Im wondering if you need to pass a user id here to get the image for this specific user.

dash102: I just did it for the current logged-in user, because our app doesn't currently have the functionality to view other profiles

dash102: I'm thinking that if we have a feature next week that involves showing profiles, this function would be updated there instead

nikhitashanker: Understood, makes sense.

yicyao:
LGTM

PR #137, opened 2020-08-17
Add interest filters to discovery page

yicyao:
Adds the interest filters corresponding to event labels to discovery page, and change filter categories to be more general.
Also changes date format to prevent wrong date format. 
ac5264a1-3a10-4ab8-aa38-d5d215896fd1
date-picker-chrome


project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
```diff
@@ -62,4 +62,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     // Event in database
     response.getWriter().println(CommonUtils.convertToJson(SpannerTasks.getEventById(event.getId()).get().toBuilder().build()));
   }
+  private static List<String> splitAsList(String values) {
```
dash102, line 65: Feel free to move this to CommonUtils because I think it's the second declaration of this function

project/src/main/webapp/events-feed.html
```diff
@@ -45,6 +45,10 @@ <h3>Discover Events Around You</h3>
           aria-expanded="false"
         >
           Filter
+        <span class="caret"></span></button>
+        <ul class="dropdown-menu" id="filterCheckBoxes" role="menu" aria-labelledby="menu1">
+        <button class="filter-btn" type="submit" onclick="getFilteredEvents()">Filter</button>
```
dash102, line 50: nit: indent this button since it's inside the list

project/src/main/webapp/events-feed.js
```diff
@@ -3,6 +3,10 @@ $(async function() {
   populateAllEvents(allEvents);
 });
 
+$(document).ready(function(){
```
dash102, line 13: The function above this is already a document.ready one, so the toggle should probably either go straight in there or in a short function called `toggleDropdown` and called from the above function

project/src/main/webapp/events-feed.js
```diff
@@ -11,5 +15,48 @@ async function getAllEvents() {
 function populateAllEvents(allEvents) {
   for (const event of allEvents) {
     populateEventContainer(event, 'event-container');
+    getInterestFilters(event);
+  }
+  populateFilters(filters)
+}
+
+const filters = {}
```
dash102, line 40: This might just be personal preference, but if you declare globals, I'd group them at the top of the file (or under the document.ready function) so they don't get lost in the middle of code and decoupled from the method they're supposed to be associated with over time

project/src/main/webapp/events-feed.js
```diff
@@ -11,5 +15,48 @@ async function getAllEvents() {
...
+      filters[event.labels[interest]] = 1;
+    } else {
+      filters[event.labels[interest]] = filters[interest] + 1;
+    }
   }
 }
+
+// Populate filter tab with label checkboxes
+function populateFilters(filters) {
+    for (const key in filters) {
```
dash102, line 55: This looks like more indents than the other ones, fix?

project/src/main/webapp/events-feed.js
```diff
@@ -11,5 +15,48 @@ async function getAllEvents() {
...
+      $('#filterCheckBoxes')
+        .append(getCheckboxes(key, filters[key]))
+    }
+    $('#filterCheckBoxes').append(`<br>`)
+}
+
+// Gives HTML for checkboxes
+function getCheckboxes(key, value) {
+    return `<input type="checkbox" id=${key} name=${key} value=${key}>
```
dash102, line 64: Ditto with indents

project/src/main/webapp/events-feed.js
```diff
@@ -1,15 +1,76 @@
 $(async function() {
-  const allEvents = await getAllEvents();
-  populateAllEvents(allEvents);
+  // If want all events on discovery page 
+  if (!(window.location.href).includes('filtered=true')) {
```
nikhitashanker, line 3: You may want to check if a parameter exists here rather than using includes. I think you can use URLSearchParams.has()

yicyao: Will keep as is just in case we want to add other params for other cases and can 'else if'

liangqigong:
Hi Yichen, is this PR ready for review? Since I see there are a few unresolved comments so I am not sure about the status of the PR.

project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
yicyao: will be done when others get pushed to master

yicyao:
Yes ready for review. I resolved some comments but didn't put 'done', just pressed resolved on github. Will reply to comments now

project/src/main/webapp/events-feed.html
yicyao: done

project/src/main/webapp/events-feed.js
yicyao: done

project/src/main/webapp/events-feed.js
yicyao: done (this looks correct, fixed the other ones w 4 spaces instead of 2)

project/src/main/webapp/events-feed.js
yicyao: done

project/src/main/webapp/events-feed.js
yicyao: done

project/src/test/java/com/google/sps/EventSpannerTasksTest.java
```diff
@@ -126,6 +126,7 @@ public void testEventCreationDoPost() throws Exception {
     request.addParameter("time", TIME);
     request.addParameter("description", DESCRIPTION);
     request.addParameter("location", LOCATION);
+    request.addParameter("interests", "Tech");
```
liangqigong, line 129: Do you need to assert something equals to "Tech" below?

yicyao: done

PR #140, opened 2020-08-17
Filter Events by Interest Label (Backend)

yicyao:
[APPROVED, waiting for PR 137]

When the user specifies the labels they want to filter by in the frontend, sent to backend to return only those events. Frontend in #137 

See here: Filtered by events with "Work" label
ezgif com-video-to-gif


project/src/main/java/com/google/sps/servlets/FilteredEventServlet.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/get-filtered-events")
```
dash102, line 21: Might be personal preference, but I'd rename this from "get" to maybe just "filtered-events" so that, if there does end up being a doPost ever, it's not misleading/requiring renaming everywhere

project/src/test/java/com/google/sps/EventDetailTest.java
```diff
@@ -103,6 +107,34 @@ public void verifyGetAllEvents() throws IOException {
...
+      System.out.println("JSON conversion failed.");
+    }
+  }
+
+  @Test
+  public void verifyGetOneFilteredEvent() throws IOException {
+    Set<Event> expectedEvents = new HashSet(Arrays.asList(EVENT1));
+    Mockito.when(request.getParameter("labelParams")).thenReturn( "Work");
+
+    new FilteredEventServlet().doGet(request, response);
```
nikhitashanker, line 129: You can initialize the Servlet in setup to make it a bit cleaner.

project/src/test/java/com/google/sps/EventDetailTest.java
```diff
@@ -103,6 +107,34 @@ public void verifyGetAllEvents() throws IOException {
         CommonUtils.convertToJson(expectedEvents).trim(), stringWriter.toString().trim());
   }
 
+  @Test
+  public void verifyGetFilteredEvents() throws IOException {
+    Set<Event> expectedEvents = new HashSet(Arrays.asList(EVENT1, EVENT2));
+    Mockito.when(request.getParameter("labelParams")).thenReturn( "Work-Chess");
```
nikhitashanker, line 113: I think you should create constant for "labelParams"

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -450,4 +451,43 @@ public Void run(TransactionContext transaction) throws Exception {
   private static Mutation.WriteBuilder newInsertBuilderFromOpportunitySignup() {
     return Mutation.newInsertBuilder(OPPORTUNITY_SIGNUP_TABLE);
   }
+
+  /**
+   * Given filters for events, return events whose labels match those filters
+   *
+   * @param email of loggedIn user email
```
nikhitashanker, line 458: I think the param email is incorrectly described.

project/src/test/java/com/google/sps/EventDetailTest.java
```diff
@@ -103,6 +107,34 @@ public void verifyGetAllEvents() throws IOException {
         CommonUtils.convertToJson(expectedEvents).trim(), stringWriter.toString().trim());
   }
```
nikhitashanker, line 109: I think you should create a separate test for the FilteredEventServlet class since it seems like separate functionality from the EventDetail class.

project/src/test/java/com/google/sps/EventDetailTest.java
```diff
@@ -4,6 +4,7 @@
 import com.google.sps.data.Event;
 import com.google.sps.data.User;
 import com.google.sps.servlets.EventDetailServlet;
+import com.google.sps.servlets.FilteredEventServlet;
```
nikhitashanker, line 7: I think you should separate this into 2 tests for clarity.

project/src/main/java/com/google/sps/servlets/FilteredEventServlet.java
```diff
@@ -0,0 +1,39 @@
...
+public class FilteredEventServlet extends HttpServlet {
+  /** Returns events as specified by interest filters */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String labelParams = request.getParameter("labelParams");
+    String[] labelParamsArr = labelParams.split("-");
+    Set<Event> events = SpannerTasks.getFilteredEvents(labelParamsArr);
+
+    if (events.isEmpty()) {
```
nikhitashanker, line 30: iirc isEmpty() is a Java 11 method, use !isPresent() instead for compatibility with Java 8.

project/src/test/java/com/google/sps/EventDetailTest.java
yicyao: @liangqigong  @hlz03 My reasoning here for using the same test file is that I'm using the exact same constants and both FilteredEventServlet and EventDetailServlet are called on the event-discovery.html page, but should I seperate by servlet instead of by which HTML page will eventually call it? What is the best practice in this case?

project/src/main/java/com/google/sps/servlets/FilteredEventServlet.java
liangqigong: @yicyao It seems you resolved the comment without leaving a reply here? If you discussed this offline, please still leave a reply here for future readers.

project/src/test/java/com/google/sps/EventDetailTest.java
liangqigong: I think you should separate the test files by individual servlet instead of by which HTML page will eventually call it since the HTML page call can change in the future and you don't want to update your backend test every time your frontend code changes. 

And you can create a test helper class to store all the constants if you want.

project/src/test/java/com/google/sps/EventDetailTest.java
liangqigong: ditto here. Leave something and then resolve it.

project/src/test/java/com/google/sps/EventDetailTest.java
```diff
@@ -26,8 +28,10 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 import org.mockito.Mockito;
+import org.skyscreamer.jsonassert.JSONAssert;
 import org.springframework.mock.web.MockServletContext;
 
+
 @RunWith(JUnit4.class)
 public final class EventDetailTest {
```
liangqigong, line 36: Also test the case when the servlet returns SC_NOT_FOUND

project/src/test/java/com/google/sps/EventDetailTest.java
yicyao: done

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
yicyao: done

project/src/main/java/com/google/sps/servlets/FilteredEventServlet.java
yicyao: done

project/src/main/java/com/google/sps/servlets/FilteredEventServlet.java
yicyao: done, I didn't think of a doPost, thanks

project/src/test/java/com/google/sps/EventDetailTest.java
yicyao: That makes sense, thanks! Done

project/src/test/java/com/google/sps/EventDetailTest.java
yicyao: The servlet setup annotation is @BeforeClass not @Before, so I think it doesnt apply

project/src/test/java/com/google/sps/EventDetailTest.java
yicyao: done

project/src/test/java/com/google/sps/EventDetailTest.java
yicyao: done

PR #143, opened 2020-08-17
[1/X] Add servlet to generate search results using naive algorithm.

nikhitashanker:
- Use naive algorithm to generate search results
- Add tests

### DIFF
--- a/project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
@@ -1,17 +1,3 @@
-// Copyright 2019 Google LLC
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     https://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
 package com.google.sps.servlets;
 
 import com.google.common.collect.ArrayListMultimap;
@@ -24,13 +10,20 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 /** Servlet that handles search queries and addition of event keywords to index. */
-@WebServlet("/searchData")
+@WebServlet("/search-data")
 public class SearchDataServlet extends HttpServlet {
   private ListMultimap<String, String> keywordToEventIds = ArrayListMultimap.create();
   private static final String KEYWORD = "keyword";
   private static final String EVENT_ID = "event-id";
   private static final String DESCRIPTION = "description";
-
+  
+  /**
+   * Return search results for the keyword specified in the request.
+   *
+   * @param request servlet request
+   * @param response servlet response
+   * @throws IOException if Input/Output error occurs
+   */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String keyword = request.getParameter(KEYWORD);
@@ -43,6 +36,13 @@ public class SearchDataServlet extends HttpServlet {
     response.getWriter().println(CommonUtils.convertToJson(keywordToEventIds.get(keyword)));
   }
 
+  /**
+   * Add keywords in the description specified to the index with mapping to the event id.
+   *
+   * @param request servlet request
+   * @param response servlet response
+   * @throws IOException if Input/Output error occurs
+   */
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String eventId = request.getParameter(EVENT_ID);
@@ -58,6 +58,9 @@ public class SearchDataServlet extends HttpServlet {
     response.sendRedirect("/events-feed.html");
   }
 
+  /**
+    * Add keywords to the index with mapping to given event ID.
+    */
   private void addKeywordsToIndex(String eventId, String[] keywords) {
     for (String keyword : keywords) {
       keywordToEventIds.put(keyword, eventId);
--- a/project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
@@ -59,8 +59,8 @@ public class SearchDataServlet extends HttpServlet {
   }
 
   /**
-    * Add keywords to the index with mapping to given event ID.
-    */
+   * Add keywords to the index with mapping to given event ID.
+   */
   private void addKeywordsToIndex(String eventId, String[] keywords) {
     for (String keyword : keywords) {
       keywordToEventIds.put(keyword, eventId);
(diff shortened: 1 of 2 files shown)


project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
```diff
@@ -0,0 +1,67 @@
...
+   * Add keywords in the description specified to the index with mapping to the event id.
+   *
+   * @param request servlet request
+   * @param response servlet response
+   * @throws IOException if Input/Output error occurs
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String eventId = request.getParameter(EVENT_ID);
+    String description = CommonUtils.getParameter(request, DESCRIPTION, "");
```
yicyao, line 49: I think you should also query in the name of the event (just combine them into the string[] keywordsInDescription on line 55) because someone could title it eg 'monthly book club' and then use completely diff words in the description like 'come read with us'

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Add keywords in the description specified to the index with mapping to the event id.
+   *
+   * @param request servlet request
+   * @param response servlet response
+   * @throws IOException if Input/Output error occurs
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
yicyao, line 47: Is this called whenever the user presses 'search' or whenever an event is created? In the first case, how do you look at all events (there seems to be no db query)? in the second case, keywordToEventIds hold all of the event details of every event?

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
```diff
@@ -0,0 +1,67 @@
...
+   * Return search results for the keyword specified in the request.
+   *
+   * @param request servlet request
+   * @param response servlet response
+   * @throws IOException if Input/Output error occurs
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String keyword = request.getParameter(KEYWORD);
+    if (keyword == null) {
```
yicyao, line 30: maybe should just redirect to the main event discovery page? because in this case, wouldn't the user not be searching for anything

project/src/test/java/com/google/sps/SearchDataServletTest.java
```diff
@@ -0,0 +1,146 @@
...
+    searchDataServlet.doGet(getRequest, getResponse);
+
+    Assert.assertEquals(
+        CommonUtils.convertToJson(Arrays.asList(EVENT_ID_1)).trim(),
+        stringWriter.toString().trim());
+  }
+
+  @Test
+  public void addTwoEventsAndRetrieveResultsForKeywordInDescriptionOfTwoEvents_twoResultsReturned()
+      throws IOException {
```
yicyao, line 91: So basically when there is a search with N words in the production website, there will be N post requests called? Why do you not call once?

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
```diff
@@ -0,0 +1,67 @@
...
+public class SearchDataServlet extends HttpServlet {
+  private ListMultimap<String, String> keywordToEventIds = ArrayListMultimap.create();
+  private static final String KEYWORD = "keyword";
+  private static final String EVENT_ID = "event-id";
+  private static final String DESCRIPTION = "description";
+  
+  /**
+   * Return search results for the keyword specified in the request.
+   *
+   * @param request servlet request
```
dash102, line 23: I feel like the @ part of this javadoc (not the first line) is kinda redundant because this method signature is both self-explanatory and a convention in Java already.

If this were a public-facing API I think it would make more sense to have a full doc like this, but in our case I feel like it kinda clutters the code

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
nikhitashanker: First, I will get the algorithm working, then I will add persistent storage.

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
nikhitashanker: Description will be used in improved algorithm implemented today.

project/src/test/java/com/google/sps/SearchDataServletTest.java
nikhitashanker: Here, post is supposed to happen on insertion of an event but I do not want to touch the EventCreation servlet before we have tested our MVP.

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
nikhitashanker: Here, I am returning a JSON of the search results so I do not think I should redirect to the discovery page.

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
nikhitashanker: I think the reason I am keeping it is the style guide requires javadoc for public methods to follow have @param and @return. Yes, it could possibly be cluttering the code but it can help distinguish the servlet methods from helper methods.

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
yicyao: Ok I see!

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
liangqigong, line 49: Add comment for ""

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
```diff
@@ -0,0 +1,67 @@
...
+   * @param response servlet response
+   * @throws IOException if Input/Output error occurs
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String eventId = request.getParameter(EVENT_ID);
+    String description = CommonUtils.getParameter(request, DESCRIPTION, "");
+    if (eventId == null) {
```
liangqigong, line 50: Check this early.

project/src/test/java/com/google/sps/SearchDataServletTest.java
```diff
@@ -0,0 +1,137 @@
...
+    Mockito.when(postRequest.getParameter(PARAMETER_EVENT_ID)).thenReturn(EVENT_ID_1);
+    Mockito.when(getRequest.getParameter(PARAMETER_KEYWORD)).thenReturn(CAKE);
+
+    searchDataServlet.doPost(postRequest, postResponse);
+    searchDataServlet.doGet(getRequest, getResponse);
+
+    Assert.assertEquals(CommonUtils.convertToJson(Arrays.asList()), stringWriter.toString().trim());
+  }
+
+  @Test
```
liangqigong, line 82: I think you can remove this test. The 2 results test below should be already enough if you don't have any exclusive logic for one result.

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
nikhitashanker: Done.

project/src/test/java/com/google/sps/SearchDataServletTest.java
nikhitashanker: Done - I see.

project/src/main/java/com/google/sps/servlets/SearchDataServlet.java
nikhitashanker: Done.

PR #145, opened 2020-08-18
[1/X] Event ranking (BACKEND)

dash102:
Algorithm and corresponding tests for event ranking - currently implementing a simpler ranking algorithm


project/src/main/java/com/google/sps/utilities/EventRanker.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public static List<Event> rankEvents(User user, Set<Event> unrankedEvents) {
+    List<Event> rankedEvents = new ArrayList<>(unrankedEvents);
+    Collections.sort(
+        rankedEvents,
+        (Event event1, Event event2) -> {
+          int comparison = getEventScore(user, event1).compareTo(getEventScore(user, event2));
+          if (comparison == 0) {
+            comparison = event2.getDate().compareTo(event1.getDate());
+          }
```
yicyao, line 29: maybe take out here and put in a compareTo function for code clarity

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    USER_CONSERVATION_FOOD_MUSIC =
+        TestUtils.newUser().toBuilder()
+            .setInterests(INTERESTS_CONSERVATION_FOOD)
+            .setSkills(SKILLS_MUSIC)
+            .build();
+    OPPORTUNITY_MUSIC =
+        new VolunteeringOpportunity.Builder(EVENT_FOOD_MUSIC.getId(), "", 1).build();
+    EVENT_FOOD_MUSIC = EVENT_FOOD_MUSIC.toBuilder().addOpportunity(OPPORTUNITY_MUSIC).build();
+  }
+
```
yicyao, line 69: also test behavior where user has no interests, but there are multiple events (we should probably just sort by date then)

project/src/main/java/com/google/sps/utilities/EventRanker.java
nikhitashanker, line 29: Can define a custom comparator for the Events class and use it to sort the Events here.

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,115 @@
...
+  public static void setUp() throws Exception {
+    int currentYear = new java.util.Date().getYear();
+    EVENT_CONSERVATION_FOOD_MUSIC =
+        TestUtils.newEvent().toBuilder()
+            .setLabels(new HashSet<>(Arrays.asList(CONSERVATION, FOOD, MUSIC)))
+            .build();
+    EVENT_FOOD_MUSIC =
+        TestUtils.newEvent().toBuilder()
+            .setLabels(new HashSet<>(Arrays.asList(FOOD, MUSIC)))
+            .setDate(Date.fromYearMonthDay(currentYear + 1, 1, 1))
```
nikhitashanker, line 49: This information of the year is purposely being set to create a specific set of result, so should be moved out of setup?
ex. currentYear + 1 etc

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,115 @@
...
+            Arrays.asList(
+                EVENT_FOOD, EVENT_SEWING, EVENT_CONSERVATION_FOOD_MUSIC, EVENT_FOOD_MUSIC));
+    List<Event> expectedEventRanking =
+        Arrays.asList(EVENT_CONSERVATION_FOOD_MUSIC, EVENT_FOOD_MUSIC, EVENT_FOOD, EVENT_SEWING);
+
+    List<Event> actualEventRanking =
+        EventRanker.rankEvents(USER_CONSERVATION_FOOD_MUSIC, eventsToRank);
+
+    Assert.assertEquals(expectedEventRanking, actualEventRanking);
+  }
```
nikhitashanker, line 90: I think we do not see why the expected event ranking is the ranking given just be reading the test.

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,115 @@
...
+        TestUtils.newEvent().toBuilder().setLabels(new HashSet<>(Arrays.asList(FOOD))).build();
+    EVENT_SEWING =
+        TestUtils.newEvent().toBuilder().setLabels(new HashSet<>(Arrays.asList(SEWING))).build();
+    USER_CONSERVATION_FOOD_MUSIC =
+        TestUtils.newUser().toBuilder()
+            .setInterests(INTERESTS_CONSERVATION_FOOD)
+            .setSkills(SKILLS_MUSIC)
+            .build();
+    OPPORTUNITY_MUSIC =
+        new VolunteeringOpportunity.Builder(EVENT_FOOD_MUSIC.getId(), "", 1).build();
```
nikhitashanker, line 66: nit: use TestUtils.newVolunteeringOpportunityWithEventId here.

project/src/main/java/com/google/sps/utilities/EventRanker.java
dash102: Doing this by putting it in a custom comparator for the Events class

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Done

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Done

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Done

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: It's because the user has labels of Conservation, Food, and Music, and the events need to be sorted in order of # matches: 3, 2, 1, 0

project/src/main/java/com/google/sps/utilities/EventRanker.java
dash102: I don't think I can do this because the comparator depends on something aside from the event itself, so there's no implicit ordering

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,159 @@
...
+    OPPORTUNITY_MUSIC =
+        TestUtils.newVolunteeringOpportunityWithEventId(EVENT_FOOD_MUSIC.getId());
+    EVENT_FOOD_MUSIC = EVENT_FOOD_MUSIC.toBuilder().addOpportunity(OPPORTUNITY_MUSIC).build();
+  }
+
+  @Test
+  public void testRankingEmptyEvents() throws IOException {
+    Assert.assertEquals(
+        new ArrayList<Event>(),
+        EventRanker.rankEvents(USER_CONSERVATION_FOOD_MUSIC, new HashSet<Event>()));
```
liangqigong, line 75: Try to reduce the information noise in your test. For example in this test, you just need to use a random User, not a User with food interests and music skill.

So I would use USER or newUser() here.

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,159 @@
...
+  private static Event EVENT_CONSERVATION_MUSIC;
+  private static Event EVENT_FOOD;
+  private static Event EVENT_SEWING;
+  private static User USER_CONSERVATION_FOOD_MUSIC;
+  private static User USER_NO_INTERESTS_OR_SKILLS;
+  private static VolunteeringOpportunity OPPORTUNITY_MUSIC;
+
+  @BeforeClass
+  public static void setUp() throws Exception {
+    int currentYear = new java.util.Date().getYear();
```
liangqigong, line 42: Not used?

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -0,0 +1,159 @@
...
+    EVENT_CONSERVATION_FOOD_MUSIC =
```
liangqigong, line 43: It is kinda waste here since every variable is rebuilt before running every test, but most of them are not used in a test.

Maybe create functions like newEvent(String... labels), newUser(Set interests, Set skills)

liangqigong: I would also create a function newUser(), which can call newUser(Set interests, Set skills) with empty sets.

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: That makes sense, done

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Thanks for catching! Done

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Resolving because the @BeforeClass annotation only builds the variables once

PR #144, opened 2020-08-18
Fix some test-failing code because of merge

dash102:
Merging a bunch of PRs today (from approval on Friday) caused some weird merging issues that GitHub didn't alert us of. This code fixes those.

PR #147, opened 2020-08-18
[2/X] Event ranking (BACKEND)

dash102:
Includes servlet and corresponding servlet tests for #144


project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -1,26 +1,48 @@
...
 import com.google.sps.utilities.TestUtils;
 import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import javax.servlet.ServletContextEvent;
```
yicyao, line 24: nit: goes after line 26

yicyao:
LGTM

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -112,4 +126,86 @@ public void testRankingTiedEvents() throws IOException {
...
+            .setHost(USER_CONSERVATION_FOOD_MUSIC)
+            .setLabels(new HashSet<>(Arrays.asList(FOOD)))
+            .build();
+    EVENT_SEWING =
+        TestUtils.newEvent().toBuilder()
+            .setHost(USER_CONSERVATION_FOOD_MUSIC)
+            .setLabels(new HashSet<>(Arrays.asList(SEWING)))
+            .build();
+    OPPORTUNITY_MUSIC =
+        new VolunteeringOpportunity.Builder(EVENT_FOOD_MUSIC.getId(), "", 1).build();
```
nikhitashanker, line 197: nit: can use getVolOppWithEventId from TestUtils here.

project/src/test/java/com/google/sps/EventRankerTest.java
```diff
@@ -112,4 +126,86 @@ public void testRankingTiedEvents() throws IOException {
...
+
+  @Test
+  public void testServletGetRanking() throws IOException {
+    setAuthenticationHelper(EMAIL);
+    SpannerTasks.insertOrUpdateUser(USER_CONSERVATION_FOOD_MUSIC);
+    Set<Event> events = new HashSet<>(Arrays.asList(EVENT_FOOD, EVENT_SEWING));
+    for (Event event : events) {
+      SpannerTasks.insertorUpdateEvent(event);
+    }
+    List<Event> expectedEvents = EventRanker.rankEvents(USER_CONSERVATION_FOOD_MUSIC, events);
```
nikhitashanker, line 138: This is interesting you are calling EventRanker function to get the list of expectedEvents -
I have not seen this before but if the algorithm is complicated, I can see why if might be needed.

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Looking at the linter, I think it's ok with this because lines 25 and 26 are one level deeper and are sorted after this one

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: Done

project/src/test/java/com/google/sps/EventRankerTest.java
dash102: To me, it would make sense to call EventRanker because we want the servlet to return the ranked events based on EventRanker (regardless of whether EventRanker is right or not, because that's left up to another test class).

project/src/main/java/com/google/sps/servlets/EventRankerServlet.java
```diff
@@ -0,0 +1,37 @@
+package com.google.sps.servlets;
```
liangqigong, line 1: Please also add test for this file. You might not need to test every detail of rankEvents, just test if a normal case of calling rankEvents, and a case that user is not present.

dash102: Done, moved to new file

project/src/test/java/com/google/sps/EventRankerServletTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Test
+  public void testServletGetRanking() throws IOException {
+    setAuthenticationHelper(EMAIL);
+    SpannerTasks.insertOrUpdateUser(USER);
+    Set<Event> events = new HashSet<>(Arrays.asList(EVENT_FOOD, EVENT_SEWING));
+    for (Event event : events) {
+      SpannerTasks.insertorUpdateEvent(event);
+    }
+    List<Event> expectedEvents = EventRanker.rankEvents(USER, events);
```
liangqigong, line 82: I understand why you use rankEvents here. But in general we don't want to make test to depend another code, since the test might false passed and failed due to another part of your production. Let's hardcode the expectedEvents to make the test not depend on EventRanker.

dash102: Done - that makes sense, I thought that since we tested EventRanker elsewhere we could rely on it working in this test. But it would be more reliable to hardcode.

PR #151, opened 2020-08-19
[1/Z] Event ranking

dash102:
This PR displays event cards at varying levels of detail (currently, this part is not connected to the backend and levels of detail are randomly selected from 3 presets).
Screenshot 2020-08-19 at 4 06 56 PM


project/src/main/webapp/shared.js
```diff
@@ -146,29 +152,43 @@ function togglePrefilledInterests() {
 }
 
 /** Writes out relevant details to an event card */
-async function populateEventContainer(event, containerId) {
-  const indexOfEventCard = 25;
-  const eventCardTotal = await $.get('event-card.html');
+async function populateEventContainer(event, containerId, lod = 3) {
```
nikhitashanker, line 155: Curious, what does lod stand for? Is it possible to give the longer name here or add javadoc to describe what it is?

dash102: lod stands for level of detail - I think it's a general term/acronym, but I'll document it the first time it's mentioned

project/src/main/webapp/shared.js
```diff
@@ -146,29 +152,43 @@ function togglePrefilledInterests() {
...
-  $(`#${eventCardId} #event-card-volunteers`)
-      .html(buildVolunteers(event.opportunities));
+  if (lod > 1) {
+    $(`#${eventCardId} #event-card-date`)
+        .html(
+            buildDate(
+                event.date.year, event.date.month, event.date.dayOfMonth));
+    $(`#${eventCardId} #event-card-time`).html(event.time);
+    $(`#${eventCardId} #event-card-location`).html(event.location);
+    if (event.opportunities.length > 0 ) {
```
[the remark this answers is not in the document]
dash102: Done

project/src/main/webapp/shared.js
```diff
@@ -145,30 +151,66 @@ function togglePrefilledInterests() {
...
+
+/** Prepends either a background image or Bootstrap-colored div */
+function addEventImage(imageUrl, eventCardId) {
+  if (imageUrl) {
+    $(`#${eventCardId} #event-card-image`).attr('src', imageUrl);
+  }
+  else {
+    $(`#${eventCardId} #event-card-image`).replaceWith('<div class="card-img-top" id="event-card-image" />')
+    $(`#${eventCardId} #event-card-image`).addClass(pickRandomColorClass());
+    $(`#${eventCardId} #event-card-image`).height('200px');
```
yicyao, line 198: can we just set this in css?

yicyao:
LGTM :)

project/src/main/webapp/shared.js
```diff
@@ -145,30 +151,66 @@ function togglePrefilledInterests() {
...
+  }
+}
+
+/** Pick random color from Bootstrap defaults */
+function pickRandomColorClass() {
+  const colorClasses =
+      ['bg-primary',
```
yicyao, line 205: can we take out any black/gray colors? it kinda gives the impression that it wasn't loaded correctly (that's what i initially thought in the image example)

project/src/main/webapp/shared.js
dash102: I don't think we can because it's based on logic

project/src/main/webapp/events-feed.html
```diff
@@ -63,7 +64,8 @@ <h3>Discover Events Around You</h3>
     </div>
     <br />
     <br />
-    <div class="container" id="event-container">
+    <div class="grid" id="event-container">
+      <div class="grid-sizer"></div>
```
yicyao, line 68: can we max grid size to like 4 or 5 events per row and not dynamically resize the width? like it'll always have space for x events but not change size if there are eg 3 events

project/src/main/webapp/shared.js
dash102: That's a good point, done

project/src/main/webapp/events-feed.html
dash102: Good point - the width is actually not dynamically resized, and if you don't fill up a column the size stays the same (and it wraps to the next row if it's too many events)

PR #153, opened 2020-08-19
[1/Y] Add optimized algorithm for search.

nikhitashanker:
-Add optimized algorithm described in design doc: The composite ranking of the title and the description for keyword K is computed as a weighted sum of the relevance in title and description
-Add KeywordHelper to allow testing with NLP API
-Add SearchStore as an in memory store of the index [persistent storage will be added in next PR]


nikhitashanker:
KeywordHelper was added to prevent making methods in servlet public for the purposes of testing.

project/src/main/java/com/google/sps/data/EventResult.java
```diff
@@ -0,0 +1,31 @@
+package com.google.sps.data;
+
+import java.util.Comparator;
+
+/* Class that represent an event result in search. */
+public class EventResult {
+  private String eventId;
+  private float ranking;
```
dash102, line 8: Why float? (purely for my understanding, since I kinda associate ranking with integers)

project/src/main/java/com/google/sps/store/SearchStore.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/** Store class that uses in-memory map to hold search results. */
+public class SearchStore {
+  private ListMultimap<String, EventResult> keywordToEventResults = ArrayListMultimap.create();
+  private static final float WEIGHT_IN_TITLE = 0.7f;
```
dash102, line 18: I think a comment to clarify where these weights came from/why one matters more would be helpful

project/src/main/java/com/google/sps/utilities/KeywordHelper.java
```diff
@@ -0,0 +1,53 @@
...
+
+/* Class that abstracts away getting keywords for specific content to enable testing. */
+public class KeywordHelper {
+  
+  /** Singleton instance of KeywordHelper. */
+  private static KeywordHelper instance;
+
+  /** Returns the singleton instance of KeywordHelper. */
+  public static KeywordHelper getInstance() {
+    if (instance == null) {
```
dash102, line 22: nit: Look into some shorthand for this? I only know of such a thing in python, but I found a few links like this one that might have some one-liner ideas
https://stackoverflow.com/questions/32988485/java-different-approaches-to-handle-if-null-then-new/32988645

project/src/main/java/com/google/sps/data/EventResult.java
nikhitashanker: So ranking is calculated using output from nlp library which is float value.

project/src/main/java/com/google/sps/utilities/KeywordHelper.java
nikhitashanker: Will keep in mind. Seems like writing it out long defines the purpose of the code clearly.

project/src/main/java/com/google/sps/store/SearchStore.java
nikhitashanker: Done - added javadoc to helper method and clarity to comment above function.

PR #161, opened 2020-08-20
[2/Z] Link frontend and backend events

dash102:
This PR 1) links frontend and backend of ranking and 2) does some accessory styling

Screenshot 2020-08-19 at 7 38 43 PM


yicyao:
The event cards look so beautiful! If you have time tomorrow and are working on styling/ bugs, could you make event-details.html look kinda like a giant event-card.html?

project/src/main/webapp/event-card.html
```diff
@@ -84,13 +105,12 @@ <h5 class="card-title" id="event-card-title">
...
+        <br />
         <ul class="list-group list-group-flush" id="details">
-          <button class="btn">
-            <a href="event-details.html" id="event-details">See more details</a>
-          </button>
-          <button type="button" class="btn btn-primary">
-            <a href="event-details.html" id="event-register">Register</a>
-          </button>
+          <div class="btn-group" role="group">
+            <a class="btn btn-outline-primary" href="event-details.html" id="event-details">Details</a>
```
yicyao, line 111: just want to confirm that these links are still set in a js file as something like 'event-details.html?eventId= XXX' and not just plain 'event-details.html'

project/src/main/webapp/events-feed.js
```diff
@@ -1,15 +1,20 @@
-$(async function() {
-  const allEvents = await getAllEvents();
-  populateAllEvents(allEvents);
+$(function() {
+  populateEvents();
 });
```
nikhitashanker, line 4: javadoc?

project/src/main/webapp/events-feed.js
```diff
@@ -1,15 +1,20 @@
-$(async function() {
-  const allEvents = await getAllEvents();
-  populateAllEvents(allEvents);
+$(function() {
```
nikhitashanker, line 1: notation looks a bit odd, but I am not too familiar with jquery. Is there a reason not to use onload here?

project/src/main/webapp/events-feed.html
```diff
@@ -5,6 +5,7 @@
     <title>STEP 188 Project</title>
     <link rel="stylesheet" href="events-feed.css" />
     <link rel="stylesheet" href="styles.css" />
+    <script src="https://kit.fontawesome.com/0d580893be.js" crossorigin="anonymous"></script>
```
nikhitashanker, line 8: nit: Im wondering if there is a similar google font and you can use that js file.

nikhitashanker:
LGTM besides minor fixes. UI looks really good.

project/src/main/webapp/events-feed.html
dash102: This isn't for fonts, it's for the icons

project/src/main/webapp/events-feed.js
dash102: It's the same as document.ready, just shorthand for that. It's specific to jquery

project/src/main/webapp/events-feed.js
dash102: With this function already being notated as a `document.ready` function, I don't think it would be helpful to add a comment like "function runs once document is ready" or "calls first set of functions for document"

project/src/main/webapp/event-card.html
dash102: Good catch - yup, this PR keeps those links now

project/src/main/webapp/events-feed.js
```diff
@@ -1,15 +1,20 @@
...
+  populateEvents();
 });
 
+async function populateEvents() {
+  const rankedEvents = await getRankedEvents();
+  const lodMap = getLodsFromEvents(rankedEvents);
```
liangqigong, line 7: What's lodMap? Try to use a more meaningful name if possible?

dash102: Renamed to eventLevels instead, done

PR #170, opened 2020-08-20
Add persistent storage of index.

nikhitashanker:


nikhitashanker:
Reverted SearchStore to SearchHelper because changes cannot be seen clearly in Git. I will open another small PR to do this.

project/src/main/java/com/google/sps/store/SearchStore.java
```diff
@@ -59,27 +66,22 @@ private void addKeywordsInTextToIndex(String eventId, String text, float weight)
 
     for (Keyword keyword : keywords) {
       float keywordRank = 0;
```
dash102, line 68: nit: For consistency, if these two ranking variables are referring to the same kind of ranking, either make them both primitive or both objects?

project/src/main/java/com/google/sps/store/SearchStore.java
```diff
@@ -59,27 +66,22 @@ private void addKeywordsInTextToIndex(String eventId, String text, float weight)
...
-      if (existingResult.isPresent()) {
-        keywordRank += existingResult.get().getRanking();
-        keywordToEventResults.remove(keyword.getName(), existingResult.get());
+      Float existingRanking = keywordToRanking.get(keyword.getName());
+      if (existingRanking != null) {
+        keywordRank += existingRanking;
       }
-      keywordRank += weight * keyword.getRelevance();
-      keywordToEventResults.put(keyword.getName(), new EventResult(eventId, keywordRank));
+      keywordRank = Float.sum(keywordRank, weight * keyword.getRelevance());
```
dash102, line 73: nit: Put this in a calculatedWeightedRank type of function? Might help with readability and also if this procedure takes multiple lines in the future

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -600,4 +606,84 @@ public Void run(TransactionContext transaction) throws Exception {
...
+    }
+  }
+
+  /**
+   * Returns events by keyword in descending order of ranking.
+   *
+   * @param keyword keyword to fetch event results for
+   * @return list of events in descending order of ranking
+   */
+  public static List<Event> getEventResultsByKeywordWithDescendingRanking(String keyword) {
```
dash102, line 659: In this query, add this line to only get events after today (I have a PR that does this for the other old functions):
`" AND DATE_DIFF(Date, CURRENT_DATE(), DAY) > 0"`

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -600,4 +606,84 @@ public Void run(TransactionContext transaction) throws Exception {
...
+                          .to(keywordId)
+                          .set(EVENT_ID)
+                          .to(eventId)
+                          .set(RANKING)
+                          .to(ranking)
+                          .build());
+                  return null;
+                }
+            });
+    }
```
yicyao, line 650: nit: consistent spacing? seems to be a combo of 2 or 4 spaces

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -600,4 +606,84 @@ public Void run(TransactionContext transaction) throws Exception {
...
+  public static List<Event> getEventResultsByKeywordWithDescendingRanking(String keyword) {
+    List<Event> results = new ArrayList<Event>();
+    Statement statement =
+        Statement.of(
+            String.format(
+                "SELECT Events.EventID, Events.Name, Events.Description, Events.Labels,"
+                    + " Events.Location, Events.Date, Events.Time, Events.Host FROM Results INNER"
+                    + " JOIN Keywords ON Results.KeywordID = Keywords.KeywordID INNER JOIN Events"
+                    + " ON Events.EventID = Results.EventID WHERE Keywords.Name=\"%s\" ORDER BY"
+                    + " Results.Ranking DESC LIMIT 20;",
```
yicyao, line 668: is there a reason why there's a limit to # of events returned?

project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -92,166 +94,237 @@
...
     mockKeywordHelper = Mockito.mock(KeywordHelper.class);
     searchStore = new SearchStore(mockKeywordHelper);
   }
 
+  @After
+  public void tearDown() throws Exception {
+    SpannerTestTasks.cleanup();
+  }
+
+
```
yicyao, line 118: nit: extra space

project/src/main/java/com/google/sps/store/SearchStore.java
```diff
@@ -59,27 +66,22 @@ private void addKeywordsInTextToIndex(String eventId, String text, float weight)
...
+   * Gets event searchResults for the given keyword search.
+   *
+   * @param keyword keyword to search
+   * @return list of events in descending order of ranking as search result
    */
-  public List<String> getSearchResults(String keyword) {
-    List<EventResult> results = keywordToEventResults.get(keyword);
-    Collections.sort(results, EventResult.ORDER_BY_RANKING_DESC);
-    return results.stream().map(EventResult::getEventId).collect(Collectors.toList());
+  public static List<Event> getSearchResults(String keyword) {
```
yicyao, line 84: Looking ahead, you'll probably also want to call part of Stephanie's function to rank the events (so you would return all of the correct events, but they're also in a ranked order)

nikhitashanker: I see the thinking but for user manual search, there is a different ranking algorithm.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
nikhitashanker: Copy pasted this from linter.

project/src/test/java/com/google/sps/SearchStoreTest.java
nikhitashanker: Done.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
nikhitashanker: I am planning on showing a maximum of 20 results for a search.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
nikhitashanker: Done - thank you.

project/src/main/java/com/google/sps/store/SearchStore.java
nikhitashanker: Done.

nikhitashanker: Done - yes I agree.

project/src/main/java/com/google/sps/store/SearchStore.java
nikhitashanker: I removed the use of Float.sum so it is very small amount of code. Will keep in mind for future but no plans to change algorithm for now.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -600,4 +606,88 @@ public Void run(TransactionContext transaction) throws Exception {
...
+              new TransactionCallable<Void>() {
+                @Override
+                public Void run(TransactionContext transaction) throws Exception {
+                  String eventId = result.getEventId();
+                  String keyword = result.getKeyword();
+                  float ranking = result.getRanking();
+                  Struct row =
+                      transaction.readRow(
+                          KEYWORDS_TABLE, Key.of(keyword), Collections.singleton(NAME));
+                  String keywordId = row == null ? UUID.randomUUID().toString() : row.getString(0);
```
liangqigong, line 630: Put this in the if (row == null) block?

project/src/main/java/com/google/sps/utilities/TestUtils.java
```diff
@@ -19,7 +19,9 @@
   private static final String DESCRIPTION = "Daily Team Sync";
   private static final Set<String> LABELS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList("Tech", "Work")));
+  private static final Date PAST_DATE = Date.fromYearMonthDay(1999, 9, 15);
   private static final String LOCATION = "Remote";
+  private static final Date FUTURE_DATE = Date.fromYearMonthDay(2021, 9, 15);
```
liangqigong, line 24: Can you use may currentDate() + 1 year? Since 2021 could be past in the future.

project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -92,166 +94,250 @@
...
+    // On the first call the getKeywords for name, return empty list
     // On the second call the getKeyords for description, return empty list
     Mockito.when(mockKeywordHelper.getKeywords())
-        .thenReturn(KEYWORDS_TITLE_WITHOUT_GAMES, KEYWORDS_DESCRIPTION_WITHOUT_GAMES);
+        .thenReturn(KEYWORDS_NAME_WITHOUT_GAMES, KEYWORDS_DESCRIPTION_WITHOUT_GAMES);
     
-    searchStore.addEventToIndex(EVENT_ID_1, TITLE_WITHOUT_GAMES, DESCRIPTION_WITHOUT_GAMES);
-    List<String> actualResults = searchStore.getSearchResults(GAMES);
+    searchStore.addEventToIndex(EVENT_ID_1, NAME_WITHOUT_GAMES, DESCRIPTION_WITHOUT_GAMES);
+    List<Event> actualResults = searchStore.getSearchResults(GAMES);
```
liangqigong, line 131: Usually we don't call 2 functions in one test because the test can pass when none of these functions work. For example, addEventToIndex doesn't work properly so nothing is added, getSearchResults doesn't work properly but since there is nothing in the DB, so the test passes.

I understand that you write it this way because you have no other way to add Event to your DB. So I would recommend you to add a comment for the test to describe it tests 2 behaviors. And also Assert if the Event is added to DB before  getSearchResults

project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -92,166 +94,250 @@
...
+    // 1                     Yes                    No
+    SpannerTasks.insertorUpdateEvent(
+        TestUtils.newEventWithFutureDate(EVENT_ID_1, NAME_WITH_GAMES, DESCRIPTION_WITHOUT_GAMES));
+    Mockito.when(mockKeywordHelper.getKeywords())
+        .thenReturn(KEYWORDS_NAME_WITH_GAMES, KEYWORDS_DESCRIPTION_WITHOUT_GAMES);
+
+    searchStore.addEventToIndex(EVENT_ID_1, NAME_WITH_GAMES, DESCRIPTION_WITHOUT_GAMES);
+    List<Event> actualResults = searchStore.getSearchResults(GAMES);
+
+    Assert.assertEquals(EVENT_ID_1, actualResults.get(0).getId());
```
liangqigong, line 162: Can you directly compare the classes? Try to build one Event if you can.

PR #174, opened 2020-08-20
Constrain event query results to the events after today

dash102:


yicyao:
LGTM!

project/src/test/java/com/google/sps/EventSpannerTasksTest.java
```diff
@@ -205,11 +209,29 @@ public void testEventRegistrationDoGet() throws Exception {
...
         new ReflectionEquals(returnedEvent.getAttendees(),/*excludeFields= */ null)
             .matches(new HashSet<User>(Arrays.asList(ATTENDEE, HOST))));
   }
-}
 
+  /**
+   * Verify that events in the past are NOT retrieved
+   */
+  @Test
+  public void testGetFutureEventsOnly() throws Exception {
```
nikhitashanker, line 223: you can include the result in the test name as opposed to writing a comment.

nikhitashanker:
LGTM

project/src/test/java/com/google/sps/EventSpannerTasksTest.java
dash102: That makes sense, done

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -128,7 +128,8 @@ public static void insertorUpdateEvent(Event event) {
                 Statement.of(
                     String.format(
                         "SELECT EventID, Name, Description, Labels, Location, Date, Time,"
-                            + " Host, Opportunities, Attendees, Image FROM %s WHERE EventID in (%s)",
+                            + " Host, Opportunities, Attendees FROM %s WHERE EventID in (%s)"
+                            + " AND DATE_DIFF(Date, CURRENT_DATE(), DAY) > 0",
```
liangqigong, line 132: I can think of one improvement, you don't have to make it in this PR.

You can create a AutoValue, which includes all the possible conditions. And use the AutoValue to control what condition to add in the SQL, so you don't have to rewrite the condition every time.

For example

EventTableConditionParams {
  Optional<List> eventIds;
  Optional<boolean> isFuture;
}

function generateCondition(EventTableConditionParams params) {
  condition = ""
  params.eventIds.ifPresent(condition + EventID in (%s));
  params.isFuture.ifPresent(condtion + DATE_DIFF(Date, CURRENT_DATE(), DAY) > 0)
}

PR #179, opened 2020-08-21
Configure edit and register buttons for hosts and non-hosts

dash102:
image

Fixes issues #149 and #163

### DIFF
--- a/project/src/main/webapp/event-details.js
+++ b/project/src/main/webapp/event-details.js
@@ -1,11 +1,26 @@
 window.onload = async function onLoad() {
   getEventDetails();
   const eventHost = await getEventHost();
+  configureRegisterAndEditButtons(eventHost);
   const loginStatus = await getLoginStatus();
   populateVolunteeringOpportunitiesUI(eventHost, loginStatus);
   showCreateOpportunityLink(eventHost, loginStatus);
   setSignupAction();
 };
+
+/** 
+ * Adds the edit button for the host and removes the register button
+ * @param {Object} eventHost email of the host of the event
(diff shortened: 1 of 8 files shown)


project/src/main/webapp/event-details.js
```diff
@@ -1,12 +1,27 @@
...
   const eventHost = await getEventHost();
+  configureRegisterAndEditButtons(eventHost);
   const loginStatus = await getLoginStatus();
   populateVolunteeringOpportunitiesUI(eventHost, loginStatus);
   showCreateOpportunityLink(eventHost, loginStatus);
   setSignupAction();
 };
 
+/** 
+ * Adds the edit button for the host and removes the register button
```
yicyao, line 12: If this removes the register button for the host, in what case would you get the alert for 'you cant register bc you are the host'? is it if the user manually navigates to the url?

project/src/main/webapp/event-details.js
```diff
@@ -1,12 +1,27 @@
...
+ * @param {Object} eventHost email of the host of the event
+ * @param {Object} loginStatus status of current user
```
yicyao, line 14: seems to not exist

project/src/main/webapp/event-details.js
dash102: They're 2 separate pages - the alert shows up when you click on register in the feed, and once you click into event details, it hides that register button

project/src/main/webapp/event-details.js
dash102: Oops, didn't update that - fixed!

PR #184, opened 2020-08-21
Find Event relations with label mapping

yicyao:
(550 lines are txt file, only ~400 real lines of code for review) 

Creates label mapping (and related helper functions) to determine if events are related via direct match or similar match (and by how many matches) by looking at their labels. A direct match is where the labels are the same ('basketball = basketball'), and similar match is where labels are related because in the same category/subcategory ('baseball' and 'basketball' are related because both 'sports'). 
This will allow events that are related in some fashion to show up higher in the event ranking. Eg if user likes 'baseball', we want to suggest all events related to 'baseball' (aka other sports events) before we show events that are related to other categories. Without this PR, a user who likes 'baseball' will be recommended  'cooking', 'fashion', 'basketball' all with the same weight ( with no preference to 'basketball')

I create two data structures subcategoryToCategoryMapping and categorytoSubcategoryMapping which act similar to a bi directional map to hold the subcategories of a category and vice versa. (did not use bidi map because a category has multiple things in it, so if bidi map, you can't query by subcategory key because it'll be in a set<String> not just a String). 
Example of what maps look like:
```
 (categorytoSubcategoryMapping:  Sports -> [Basketball, Soccer, Swimming] )
(subcategoryToCategoryMapping :  Basketball -> Sports, Soccer -> Sports, Swimming -> Sports
```
[This category/subcategory structure is based on https://cloud.google.com/natural-language/docs/categories, which is how the NLP API would label the events, and these are where we took the preselected labels you can choose from (for label uniformity througout). I put into txt file because there is no way to get all categories via API.]

Later Usage: 
- Event ranker takes directMatchEvents and similarMatchEvents and applies an algo to score and rank them collectively
    ArrayList<Set<Pair<Event, Integer>>> matches = labelCategories.getEventRelevancy(Spanner.getAllEvents(), user);
    Set<Pair<Event, Integer>> directMatches = matches.get(0);
    Set<Pair<Event, Integer>> similarMatches = matches.get(1);


project/src/test/java/com/google/sps/GetLabelCategoriesTest.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.HashSet;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import java.io.FileNotFoundException;
+import org.javatuples.Pair; 
+
+/* Test getEventRelevancy(), finding direct and similar matches  */
+@RunWith(JUnit4.class)
+public final class GetLabelCategoriesTest {
+  GetLabelCategories labelCategories = new GetLabelCategories();
```
nikhitashanker, line 20: private?

project/src/test/java/com/google/sps/GetLabelCategoriesTest.java
```diff
@@ -0,0 +1,170 @@
...
+      new HashSet<>(Arrays.asList(new Pair<Event, Integer>(event, /*common labels=*/ 1)));
+    Set<Pair<Event, Integer>> expectedSimilarMatches = 
+      new HashSet<>(Arrays.asList(new Pair<Event, Integer>(event, /*common labels=*/ 1),
+      new Pair<Event, Integer>(event2, /*common labels=*/ 2), new Pair<Event, Integer>(event3, /*common labels=*/ 1)));
+
+    ArrayList<Set<Pair<Event, Integer>>> matches = labelCategories.getEventRelevancy(
+      new HashSet<>(Arrays.asList(event, event2, event3, event4)), user);
+    Set<Pair<Event, Integer>> directMatches = matches.get(0);
+    Set<Pair<Event, Integer>> similarMatches = matches.get(1);
+
```
nikhitashanker, line 166: I think you can inline these two lines since variables are only used once in assert.

project/src/main/java/com/google/sps/utilities/GetLabelCategories.java
```diff
@@ -0,0 +1,210 @@
...
+    directAndSimilarMatches.add(similarLabelEvents);
+    return directAndSimilarMatches;
+  }
+
+  // Checks to see if direct matches exist, if so, add to directMatchEvents
+  private void checkDirectMatches(
+      Set<Pair<Event, Integer>> directMatchEvents, Event event,
+      Set<String>  interestsAndSkills, Set<String>  labelsAndSkills) {
+    Optional<Pair<Event, Integer>> directMatchEventsOptional = 
+    getDirectMatchEvents(event, interestsAndSkills, labelsAndSkills );
```
nikhitashanker, line 72: nit: space before );

yicyao: done

project/src/main/java/com/google/sps/utilities/GetLabelCategories.java
```diff
@@ -0,0 +1,185 @@
...
+	      }
+      }
+    }
+    catch (FileNotFoundException e) {
+      System.out.println("File not found");
+    }
+    return valueToKeyMap;
+  }
+
+  /** Creates Map<String, HashSet<String>>  with Key: Category, Value: Set(Subcategories)
```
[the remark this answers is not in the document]
yicyao: initially wanted to make it extra clear about k/v but changed the comment

project/src/test/java/com/google/sps/GetLabelCategoriesTest.java
yicyao: yes ty

project/src/test/java/com/google/sps/GetLabelCategoriesTest.java
yicyao: inlined w comments for all, done

PR #191, opened 2020-08-22
Fix event details bug

dash102:
This PR:
1) Removes the hardcoded events from event ranker
2) That in turn helped fix the bugs where event details header wasn't loaded and labels were being loaded doubly


yicyao:
LGTM

PR #194, opened 2020-08-24
[2/Y] Add search UI.

nikhitashanker:
- Use event cards uniform with events feed
- Add search bar at the top of discover events
- Ranking decreases from left to right and top to down
image

### DIFF
--- a/project/src/main/webapp/events-feed.js
+++ b/project/src/main/webapp/events-feed.js
@@ -28,10 +28,6 @@ $(async function() {
  * the events in the display.
  */
 async function getAllSearchEvents() {
-  const keyword = document.getElementById('keyword').value;
-  const response = await fetch(`search-data?keyword=${keyword}`);
-  const searchEvents = await response.json();
-
   const eventContainer = document.getElementById('event-container');
   if (eventContainer != null) {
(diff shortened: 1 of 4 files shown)


project/src/main/webapp/events-feed.js
```diff
@@ -23,6 +23,29 @@ $(async function() {
...
+
+  const searchContainer = document.getElementById('search-container');
+  searchContainer.innerHTML = '';
+
+  const keyword = document.getElementById('keyword').value;
+  const response = await fetch(`search-data?keyword=${keyword}`);
+  const searchEvents = await response.json();
+  for (const key in searchEvents) {
+    if (searchEvents.hasOwnProperty(key)) {
+      populateEventContainer(searchEvents[key], 'search-container', 4);
```
dash102, line 44: What's the difference between the old level 3 and this new level 4?

Otherwise, LGTM

nikhitashanker: In level 4, the cards are placed into strict rows to demonstrate search ranking rather than the puzzling effect.

yicyao:
LGTM

PR #199, opened 2020-08-24
[2/Y] Add words within keywords and basic forms of keywords to index.

nikhitashanker:
- Create a utility in KeywordHelper to map words to their basic forms given piece of text
- Add words within keywords and basic forms of keywords to index with the same weight of the keyword itself
- Add corresponding tests including test that tests ranking with words within keywords and basic forms of the keywords.


project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -38,66 +41,157 @@
...
-              new Keyword("lunch", 0.07f),
-              new Keyword(GAMES, 0.01f),
-              new Keyword("Clunie Pool", 0.03f)));
+   private static final ImmutableList<Keyword> KEYWORDS_NAME_WITHOUT_GAMES =
+      ImmutableList.of(new Keyword("picnic", 1.00f));
+  private static final ImmutableList<Keyword> KEYWORDS_NAME_WITH_GAMES =
+      ImmutableList.of(new Keyword("picnic", 0.56f), new Keyword(GAMES, 0.44f));
+  private static final ImmutableList<Keyword> KEYWORDS_DESCRIPTION_WITHOUT_GAMES =
+      ImmutableList.of(
+          new Keyword("Sutter Middle School", 0.43f),
```
dash102, line 50: Make these keywords into constants since they're used quite a few times?

project/src/main/java/com/google/sps/utilities/KeywordHelper.java
```diff
@@ -52,7 +57,33 @@ public void setContent(String content) {
...
+  /**
+   * Finds the tokens with their basic forms in the current instance's content.
+   *
+   * @return mapping of each token in current instance's content to its basic form.
+   */
+  public Map<String, String> getTokensWithBasicForm() {
+    Map<String, String> tokenToBasicForm = new HashMap<String, String>();
+    try (LanguageServiceClient language = LanguageServiceClient.create()) {
+      Document doc = Document.newBuilder().setContent(content).setType(Type.PLAIN_TEXT).build();
+      AnalyzeSyntaxRequest request =
```
dash102, line 75: nit: put lines 75 and 76 in their own function? I think if these details were replaced with a single named function, `getTokensWithBasicForm` would read really nicely and be closer to the same level of abstraction

project/src/test/java/com/google/sps/SearchStoreTest.java
nikhitashanker: Yes, some keywords are similar - I think here since the constants are the actual NLP output, it's easier to see the output without constants.

project/src/main/java/com/google/sps/utilities/KeywordHelper.java
nikhitashanker: Here, in the first method we are using AnalyzeEntityRequest and in the second method we are using AnalyzeSyntaxRequest. Do you mind clarifying what you mean by same level of abstraction?

dash102: sure! It's not really a big deal but, to me, these 2 lines make up a cohesive unit that could be abstracted away in this function so when you read this function, that part is just `getTokensList(doc)`

dash102: where `getTokensList(doc)` makes the request and makes the response and returns response.getTokensList()

nikhitashanker: Done - thanks for the clarification.

project/src/main/java/com/google/sps/store/SearchStore.java
```diff
@@ -46,24 +47,46 @@ public void addEventToIndex(String eventId, String name, String description) {
...
+  /**
+   * Adds keyword and basic form of the keyword if different than keyword to the index.
+   *
+   * @param keyword keyword to add to index
+   * @param ranking ranking of the keyword
+   * @param keywordToRanking the index to update
+   * @param tokenToBasicForm map of token to basic form of token for all tokens
+   *        in text where keyword was found
+   */
+  private void addKeywordAndBasicFormIfDifferent(String keyword, float ranking,
```
yicyao, line 83: i think it doesnt need to be named with 'IfDifferent' because there is no 'ifSame' version

project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -38,66 +41,156 @@
...
+      + "Event hours are 10am to 1:30pm.";
+  private static final ImmutableList<Keyword> KEYWORDS_DESCRIPTION_WITH_GROWERS =
+      ImmutableList.of(
+          new Keyword("farmers' market", 0.36f),
+          new Keyword("food vendors", 0.26f),
+          new Keyword("growers", 0.26f),
+          new Keyword("Event", 0.12f));
+  private static final ImmutableMap<String, String> TOKENS_NAME_WITHOUT_GROWERS =
+      ImmutableMap.<String, String>builder()
+          .put("wednesday", "wednesday")
```
yicyao, line 152: why do you put key: value as the same thing? what does this test

yicyao: also is this not the same constant as TOKENS_NAME_WITHOUT_FOOD_VENDORS?

project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -38,66 +41,156 @@
...
+          .put("cesar", "cesar")
+          .put("chavez", "chavez")
+          .put("farmers", "farmer")
+          .put("'", "'")
+          .put("market", "market")
+          .build();
+  private static final ImmutableMap<String, String> TOKENS_DESCRIPTION_WITH_GROWERS =
```
yicyao, line 159: same as TOKENS_DESCRIPTION_WITH_FOOD_VENDORS ?

project/src/main/java/com/google/sps/store/SearchStore.java
nikhitashanker: Done.

project/src/test/java/com/google/sps/SearchStoreTest.java
nikhitashanker: Done. Used putAll from ImmutableMap to create a copy of first map.

project/src/test/java/com/google/sps/SearchStoreTest.java
nikhitashanker: The basic form of a word may be itself I added this in the mock comments. Yes - thanks I have removed all duplicate constants using addAll of ImmutableList and putAll of ImmutableMap.

project/src/test/java/com/google/sps/SearchStoreTest.java
```diff
@@ -353,4 +424,116 @@ public void twoEvents_tieInRelevanceInDescriptionOfBothEvents_twoResultsReturned
...
+
+  /**
+   * When an event with description with the keyword food vendors is added to the index,
+   * a search for the word food should also return the event in result. Food is a 
+   * word within the keyword food vendors.
+   */
+  @Test
+  public void
+      addEventWithKeywordInDescription_searchForWordWithinKeyword_oneResultReturned()
+          throws IOException {
```
liangqigong, line 436: You can throws Exception here. Don't need to be so specific in the test.

nikhitashanker: Done.

PR #200, opened 2020-08-24
[1/X] Accessory UI changes

dash102:
Screenshot 2020-08-24 at 9 16 07 PM

Screenshot 2020-08-24 at 9 09 58 PM

Screenshot 2020-08-24 at 9 09 33 PM


project/src/main/webapp/event-details.css
```diff
@@ -33,3 +33,17 @@
 .alert {
   display: none;
 }
+
```
yicyao, line 36: want also img size limit (sometimes the img is huge)
img {
width: xxxx
height: xxxx
}

dash102: Discussed over chat - cropping and resizing don't really work out well here so we will stick with properly-sized images

dash102:
> I see the signup for volunteering opportunities and the edit volunteering opportunities are no longer visible and you can keep style as is for them. And I open a separate PR to style them?
> 'Select a volunteer role' is missing:
> image

Fixed

PR #202, opened 2020-08-25
Fix Bug 413 'Event URI too long'

yicyao:
Fix Bug 413 'Event URI too long' by passing in eventId instead of full event (when passing in full event, it concats everything about event (description, name, image url which sometimes leads to 8000+ char request params)


project/src/main/java/com/google/sps/servlets/EventRankerServlet.java
```diff
@@ -9,6 +9,7 @@
 import com.google.sps.utilities.SpannerTasks;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.ArrayList;
```
dash102, line 12: nit: alphabetize

project/src/main/webapp/shared.js
```diff
@@ -181,10 +181,15 @@ function togglePrefilledInterests() {
 }
 
 async function getRankedEvents(events) {
-  const response = await fetch('/event-ranker?' + new URLSearchParams({'events': JSON.stringify(events)}));
+  const eventIds = new Array(); 
+  for (index in events) {
```
dash102, line 185: Use the for/of loop instead, for things that are arrays 
https://google.github.io/styleguide/jsguide.html#features-for-loops

project/src/main/java/com/google/sps/servlets/EventRankerServlet.java
yicyao: done

project/src/main/webapp/shared.js
yicyao: yes ty i was looking for something like that!

PR #204, opened 2020-08-26
Fix events hosting, events participating, events volunteering bug

nikhitashanker:
- Update query for events hosting
- Add JS fixes for response in dict format


nikhitashanker:
Revert merge failed opening new PR.

PR #205, opened 2020-08-26
Fix events display on profile bugs.

nikhitashanker:
- Update query for events hosting due to eventsHosting column of user not being updated upon event insertion
-Update test to remove insertion into user's events hosting [.addHosting(event) ]
image

- Update front end to handle dictionary in JS
-Add roles for events volunteering in profile display
image


project/src/main/webapp/profile.js
```diff
@@ -32,24 +32,30 @@ function updateProfileEvents() {
 
 async function updateUserEventsHosting() {
   const userEventsHosting = await getUserEvents('hosting');
-  for (const eventHosting of userEventsHosting) {
-    populateEventContainer(eventHosting, 'events-hosting');
+  for (const eventsKey in userEventsHosting) {
```
dash102, line 35: Are you still able to use for/of here like before? If so, that seems to be more aligned w/ the JS style guide

Otherwise, LGTM!

nikhitashanker:
Converting to draft to see if I can add the roles for events volunteering.

project/src/main/webapp/profile.js
nikhitashanker: Unable to because unfortunately the data is in dictionary style.

project/src/main/webapp/shared.js
```diff
@@ -249,6 +256,8 @@ async function populateEventContainer(event, containerId, lod) {
   addLinkToDetails(eventCardId);
 }
 
+
+
```
liangqigong, line 260: Remove the empty lines?

nikhitashanker: Done.

PR #227, opened 2020-08-27
Change event ordering to left-right instead of top-bottom

dash102:
Event cards used to be ordered like the left, but this PR makes them to be displayed like the right.

image


project/src/main/webapp/events-feed.js
```diff
@@ -50,10 +50,26 @@ async function getAllEvents() {
...
+  for (let i = 0; i < 4; i++) {
+    for (const eventMap of transposedEventLevels[i]) {
+      populateEventContainer(eventMap['event'], `masonry-col-${i + 1}`, eventMap['lod']);
+      getInterestFilters(eventMap['event']);
+    }
+  }
+}
+
+/**
+ * Makes the masonry grid for events-feed ordered rougly from left to right
```
yicyao, line 63: nit: roughly

yicyao:
LGTM

project/src/main/webapp/events-feed.js
dash102: Thanks for catching 😅

PR #232, opened 2020-08-28
[1/X] Fix event update.

nikhitashanker:
- Currently event ID is not being passed into the mutation so the event is not being updated.
- If eventID is null then we insert event otherwise we update

Must be reviewed by @yicyao before merge

### DIFF
--- a/project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
+++ b/project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
@@ -74,6 +74,7 @@ public class EventCreationServlet extends HttpServlet {
   /** Posts newly created event to database with NLP suggested labels and redirects to page with created event details*/
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String eventId = request.getParameter("eventId");
     String name = request.getParameter("name");
     String[] parsedDate = request.getParameter("date").split("-");
     Date date =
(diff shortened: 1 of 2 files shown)


project/src/main/webapp/event-edit.js
```diff
@@ -20,4 +20,7 @@ async function getEventDetails() {
     `${data['date'].month}/${data['date'].dayOfMonth}/${data['date'].year}`;
   document.getElementById('time').value = data['time'];
   document.getElementById('location').value = data['location'];
+  const eventEditForm = document.getElementById('create-event');
+  eventEditForm.action =
+      `/create-event?eventId=${eventId}`;
```
yicyao, line 25: this will double POST -- the html says '  <form id="create-event" action="/create-event" method="POST">' on submit

project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
```diff
@@ -93,10 +94,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    SpannerTasks.insertorUpdateEvent(event);
-
-    // Add event to search index
-    searchStore.addEventToIndex(event.getId(), name, description);
+    if (eventId == null) {
+      SpannerTasks.insertorUpdateEvent(event);
+      searchStore.addEventToIndex(event.getId(), name, description);
+    } else {
+      event = event.toBuilder().setId(eventId).build();
+      SpannerTasks.insertorUpdateEvent(event);
```
yicyao, line 102: i would just inline

nikhitashanker: I tried this but on line 105 you have a line that gets the event ID of the event, so we should not inline.

project/src/main/webapp/event-edit.js
nikhitashanker: this line will change the action rather than appending to action, seems fine

PR #234, opened 2020-08-28
Add index update

nikhitashanker:
- Update index upon event update.
- Test requires interaction between inserting and updating index.


project/src/main/java/com/google/sps/servlets/EventCreationServlet.java
```diff
@@ -96,11 +96,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     Event event = new Event.Builder(name, description, labels, location, date, time, host).build();
     if (eventId == null) {
       SpannerTasks.insertorUpdateEvent(event);
-      searchStore.addEventToIndex(event.getId(), name, description);
     } else {
       event = event.toBuilder().setId(eventId).build();
       SpannerTasks.insertorUpdateEvent(event);
+      SpannerTasks.deleteIndexEntriesByEventId(eventId);
```
hlz03, line 102: Can we just update index instead?

nikhitashanker: I tried to think of a way but in the results table we have <EventID, KeywordID, Ranking> and we have to run the algorithm to get the new keywords. If the new keywords are of a different number of rows than the old one, no way to update without deleting or inserting depending on the difference.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -701,4 +701,23 @@ public Void run(TransactionContext transaction) throws Exception {
...
+   public static void deleteIndexEntriesByEventId(String eventId) {
+    SpannerClient.getDatabaseClient()
+        .readWriteTransaction()
+        .run(
+            new TransactionCallable<Void>() {
+              @Override
+              public Void run(TransactionContext transaction) throws Exception {
+                String sql = String.format("DELETE FROM Results WHERE EventID=\"%s\";", eventId);
+                long rowCount = transaction.executeUpdate(Statement.of(sql));
+                return null;
```
dash102, line 719: Is `return null` required? (don't actually know, curious)

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
```diff
@@ -701,4 +701,23 @@ public Void run(TransactionContext transaction) throws Exception {
...
+   * @param eventId eventId to delete search entries for
+   */
+   public static void deleteIndexEntriesByEventId(String eventId) {
+    SpannerClient.getDatabaseClient()
+        .readWriteTransaction()
+        .run(
+            new TransactionCallable<Void>() {
+              @Override
+              public Void run(TransactionContext transaction) throws Exception {
+                String sql = String.format("DELETE FROM Results WHERE EventID=\"%s\";", eventId);
```
dash102, line 717: I think in-lining this SQL call would be pretty reasonable here, since it's clear that the String is a SQL call.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
nikhitashanker: interesting - the return type of public Void run(TransactionContext transaction) throws Exception  is Void. https://www.baeldung.com/java-void-type seems like we can only return null.

project/src/main/java/com/google/sps/utilities/SpannerTasks.java
nikhitashanker: Done - inlined.

